PR #1, opened 2021-07-21
[SMAGENT-3090] enable ubuntu kernel builds in probe builder, and add missing script

speedyguy17:


build-probe-binaries
```diff
@@ -208,12 +208,24 @@ else
...
+	SYSDIG_TAG="${SYSDIG_TAG}-cfgprobe"
+fi
+
 function checkout_sysdig {
 	if [ -z "$HAVE_SYSDIG_CHECKOUT" ]
 	then
 		rm -rf $SYSDIG_DIR
 		mkdir -p $SYSDIG_DIR
-		git clone git@github.com:draios/agent-libs.git $SYSDIG_DIR
+		git clone git@github.com:draios/sysdig.git $SYSDIG_DIR
```
speedyguy17, line 216: this is just a push from what is in our downstream mirror....what are we actually using this for? It seems at the very least, this needs to be paramaterized

speedyguy17: looking at this, it seems this is only used when we don't provide -s option? is that correct? which we always do for automated builds. At which point, this becomes a philosophical question of which repo we want it to point at by default. I think ultimately this should have to point at the one that is relevant for oss-sysdig, but we can table that for now, I think.

build-probe-binaries
```diff
@@ -208,12 +208,24 @@ else
...
+# (namely, the ability to run cmake from the sysdig directory but
+# specify PROBE_NAME, PROBE_VERSION, and PROBE_DEVICE_NAME). Rather
+# than change what was in the tag we made modified tags x.y.z-cfgprobe
+# that include the sysdig version x.y.z but also one additional commit
+# that allows making the probe name/version/etc configurable.
+
+if [ $SYSDIG_TAG = "0.24.1" ] || [ $SYSDIG_TAG = "agent/0.85.1" ] || [ $SYSDIG_TAG = "falco/0.12.1" ]; then
+	SYSDIG_TAG="${SYSDIG_TAG}-cfgprobe"
+fi
+
```
gnosek: I dropped this chunk as I don't expect we'll be building those prehistoric versions but I can restore it if you prefer

build-probe-binaries
gnosek: Yes, this is for when you don't provide `-s`. Agreed about the philosophical question.

Once oss sysdig gets a release with falco-libs, we'll need some magic to check out the right falco libs instead (I don't think falco-libs are a submodule of oss-sysdig so we'd need to run cmake :/). Or drop the automatic checkout altogether and require `-s` in all cases.

gnosek (dismissed):

build-probe-binaries
speedyguy17: okay let me pull in the most recent version before merge.

gnosek (approved):

PR #5, opened 2021-08-12
[SSPROD-6903] Enhance readme after trying to validate ssprod-6903

alblasco:
First I gathered some notes when trying to validate SSPROD-6903
Second I suggested @gnosek to include in his PR
But, finally I agree that  probably a different PR is much better.

Please let me know your kind feedback


alblasco:
I see that finally @gnosek changed his mind, and included additional info about downloading kernel packages in README.md, so this PR is not longer needed

PR #12, opened 2021-09-24
Use repository metadata in kernel crawler

gnosek:
Instead of parsing the http server directory list, download the repo metadata and find the right set of packages for each supported kernel.

Removes dead CoreOS and Fedora Atomic distros, adds PhotonOS and Oracle Linux (without a dedicated crawler; not wired in to the probe builder script yet)

### DIFF
--- a/kernel-crawler.py
+++ b/kernel-crawler.py
@@ -346,12 +346,13 @@ class DebRepository(Repository):
             if not p.startswith('linux-headers-'):
                 continue
             release = p.replace('linux-headers-', '')
-            if 'linux-modules-{}'.format(release) in deps:
-                kernel_packages.append(p)
-                kernel_packages.append('linux-modules-{}'.format(release))
-            elif 'linux-image-{}'.format(release) in deps:
-                kernel_packages.append(p)
-                kernel_packages.append('linux-image-{}'.format(release))
+            candidates = ['linux-modules-{}', 'linux-image-{}', 'linux-image-{}-unsigned']
+            for c in candidates:
+                candidate = c.format(release)
+                if candidate in deps:
+                    kernel_packages.append(p)
+                    kernel_packages.append(candidate)
+                    break
 
         if not package_filter:
             return kernel_packages
--- a/kernel-crawler.py
+++ b/kernel-crawler.py
@@ -833,7 +833,7 @@ class Oracle8Mirror(OracleMirror):
 
 class EmptyMirror(object):
     @staticmethod
-    def get_package_urls():
+    def get_package_urls(version=''):
         return []
 
 
@@ -868,7 +868,7 @@ def main():
     try:
         distro_cls = DISTROS[sys.argv[1]]
     except IndexError:
-        print('Usage: kernel-crawler.py DISTRO', file=sys.stderr)
+        print('Usage: kernel-crawler.py DISTRO [VERSION]', file=sys.stderr)
         usage()
         sys.exit(1)
     except KeyError:
@@ -876,8 +876,13 @@ def main():
         usage()
         sys.exit(1)
 
+    try:
+        version = sys.argv[2]
+    except IndexError:
+        version = ''
+
     distro = distro_cls()
-    for url in distro.get_package_urls():
+    for url in distro.get_package_urls(version):
         print(url)
 
 
--- a/kernel-crawler.py
+++ b/kernel-crawler.py
@@ -345,6 +345,33 @@ class DebRepository(Repository):
         for p in deps.keys():
             if not p.startswith('linux-headers-'):
                 continue
+            # historically, we haven't built these variants
+            # and there's a ton of them. Leave them disabled,
+            # we can enable them when we need to
+            if '-cloud' in p:
+                continue
+            if '-rt' in p:
+                continue
+            if '-lowlatency' in p:
+                continue
+            if '-azure' in p:
+                continue
+            if '-oem' in p:
+                continue
+            if '-gcp' in p:
+                continue
+            if '-gke' in p:
+                continue
+            if '-oracle' in p:
+                continue
+            if '-kvm' in p:
+                continue
+            # skip backported kernels (again, to match historic behavior
+            # and to avoid an explosion in the number of built probes)
+            if '-lts-' in deps[p]['URL']:
+                continue
+            if '-hwe' in deps[p]['URL']:
+                continue
             release = p.replace('linux-headers-', '')
             candidates = ['linux-modules-{}', 'linux-image-{}', 'linux-image-{}-unsigned']
             for c in candidates:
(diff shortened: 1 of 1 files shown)


kernel-crawler.py
```diff
@@ -1,461 +1,751 @@
...
+def check_url(url):
+    resp = urlopen(url)
+    return resp.getcode() == 200
+
+
+def check_any(urls):
+    for url in urls:
+        if check_url(url):
+            return True
+    return False
```
deepskyblue86:
```suggestion
    any(check_url(url) for url in urls)
```

kernel-crawler.py
```diff
@@ -1,461 +1,751 @@
...
+            depends = current_package.get('Depends', [])
+            packages[name] = {
+                'Depends': set(depends),
+                'Version': current_package['Version'],
+                'Filename': current_package['Filename'],
+            }
+
+        return packages
+
+    KERNEL_PACKAGE_PATTERN = re.compile(r'^linux-.*?-[0-9]\.[0-9]+\.[0-9]+')
```
deepskyblue86, line 266:
```suggestion
    KERNEL_PACKAGE_PATTERN = re.compile(r'^linux-.*?-[0-9]+\.[0-9]+\.[0-9]+')
```
just to be future-proof :)

kernel-crawler.py
```diff
@@ -1,461 +1,751 @@
...
+        repo_urls = set()
+        for r in self.AL2_REPOS:
+            repo_urls.add(get_al_repo("http://amazonlinux.us-east-1.amazonaws.com/2/", r + '/x86_64'))
+        return [RpmRepository(url) for url in sorted(repo_urls)]
+
+
+class PhotonOsRepository(RpmRepository):
+    @classmethod
+    def kernel_package_query(cls):
+        return '''((name = 'linux' OR name LIKE 'linux-%devel%') AND name NOT LIKE '%esx%')'''
```
deepskyblue86, line 765: Could you please put a comment saying esx is not supported since it lacks CONFIG_TRACEPOINTS?
Maybe referencing this https://github.com/vmware/photon/issues/1223#issuecomment-909873604 too

kernel-crawler.py
```diff
@@ -1,461 +1,751 @@
...
+    OL8_REPOS = [
+        'http://yum.oracle.com/repo/OracleLinux/OL8/baseos/latest/x86_64/',
+        'http://yum.oracle.com/repo/OracleLinux/OL8/UEKR6/x86_64/',
+    ]
+
+    def __init__(self):
+        super(Oracle8Mirror, self).__init__([])
+
+    def list_repos(self):
+        return [OracleRepository(url) for url in self.OL8_REPOS]
```
deepskyblue86: These classes are the same, except for the repo list. Any reason not to have a parent class?

kernel-crawler.py
gnosek: Yup, except for the missing `return` :)

kernel-crawler.py
gnosek: I don't have an opinion on this. Do you believe this code will survive until Linux 10.0?

kernel-crawler.py
gnosek: No reason at all, I'll see if we can win something this way.

kernel-crawler.py
```diff
@@ -1,461 +1,737 @@
...
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
-
-# Author: Samuele Pilleri
-# Date: August 17th, 2015
-
-from __future__ import unicode_literals
```
speedyguy17, line 55: high level comment indicating how this works. especially for python, exceptionally helpful for people who have to debug this moving forward.

speedyguy17 (dismissed):

kernel-crawler.py
deepskyblue86: It depends on whether they decide to change their versioning speed or not. Some years ago I supposed 2.x would have lived forever.
Same if you think about Firefox, we're about to hit version 100. I would never have imagined that!

gnosek: In some way that change has already happened after Linus got bored of 2.6.lots and we're up to 5.something now ;)

kernel-crawler.py
```diff
@@ -15,447 +15,875 @@
...
+        base_query, args = cls.build_base_query(version)
+        query = '''WITH RECURSIVE transitive_deps(version, pkgkey) AS (
+                {}
+                UNION
+                SELECT transitive_deps.version, provides.pkgkey
+                    FROM provides
+                    INNER JOIN requires USING (name, flags, epoch, version, "release")
+                    INNER JOIN transitive_deps ON requires.pkgkey = transitive_deps.pkgkey
+            ) SELECT transitive_deps.version, location_href FROM packages INNER JOIN transitive_deps using(pkgkey);
+        '''.format(base_query)
```
deepskyblue86, line 629: :scream:

deepskyblue86 (dismissed):

kernel-crawler.py
```diff
@@ -15,447 +15,875 @@
...
+    from lzma import decompress as lzma_decompress
+except ImportError:
+    try:
+        from backports.lzma import decompress as lzma_decompress
+    except ImportError:
+        def lzma_decompress(content):
+            raise NotImplementedError("LZMA compression not supported, install backports.lzma")
+
+
+def get_url(url):
```
nathan-b, line 102: Shouldn't we be looking at the content-type instead of the end of the URL? This function could run into trouble if we (for example) request a .gz file but get a 404 page back or something.

kernel-crawler.py
```diff
@@ -15,447 +15,875 @@
...
+        if dependencies is None:
+            dependencies = {pkg_name}
+        pkg_deps = cls.filter_kernel_packages(packages[pkg_name]['Depends'])
+        for dep in pkg_deps:
+            dep = dep.split(None, 1)[0]
+            # Note: this always takes the first branch of alternative
+            # dependencies like 'foo|bar'. In the kernel crawler, we don't care
+            #
+            # also, apparently libc6 and libgcc1 depend on each other
+            # so we only filter for kernel packages
```
nathan-b: Should we have some logic to check the `level` param just in case we find a repository with a dependency loop in kernel packages? 

(I don't really trust repository maintainers to do their jobs correctly, in case that hasn't come up in discussion earlier ;) )

nathan-b (dismissed):

nathan-b:
Just a couple of minor quibbles; feel free to ignore if not helpful :)

kernel-crawler.py
gnosek: Hmm. We're not worried about 404s since we get an exception for them. The question is which sort of misconfiguration is more likely to happen:
- serving some random junk with the right content-type
- serving the right content with some bad content-type (like application/octet-stream)

kernel-crawler.py
gnosek: That comment is sort of outdated since we should handle circular dependencies properly now. We only filter for kernel packages because that's what we care about (otherwise you get half the debian userspace pulled in). I'll update the comment.

nathan-b (approved):

PR #13, opened 2021-09-24
List eBPF probes as well in the directory listing

gnosek:
We want the bpf .o files to show up in the directory index as well.


balinaveen (approved):
Looks good.

PR #18, opened 2021-10-18
Rewrite the probe builder in Python

gnosek:
This requires either a `pip install` from the probe builder director or (more likely) a Docker container.

The CLI is slightly changed: to actually build the probes, you need to add the `build` keyword before the command line arguments.

To run the kernel crawler, run the builder with `crawl <distro>`.

The Docker entrypoint is adapted and usage in a container should not require any changes.


probe_builder/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+def crawl(distro, version=None):
+    if version is None:
+        args = [sys.argv[0], distro]
+    else:
+        args = [sys.argv[0], distro, version]
+    kernel_crawler.main(args)
+
+
+if __name__ == '__main__':
+    cli()
```
deepskyblue86: github says "no newline at end of file"

probe_builder/kernel_crawler/__init__.py
```diff
@@ -380,82 +380,87 @@ def process_atomic_distro(current_repos):
...
 
 
-#
-# Print URLs to stdout
-#
-for url in urls:
-    if version_filter in url:
-        print(url)
+if __name__ == '__main__':
+    main(sys.argv)
```
deepskyblue86: missing newline

deepskyblue86 (dismissed):

iurly (dismissed):
Apart from newline issues, LGTM!

irozzo-1A (dismissed):
LGTM

main-builder-entrypoint.sh
```diff
@@ -108,9 +127,15 @@ done
 shift $((OPTIND - 1))
 
 case "${OP:-}" in
+	download_from_articfactory)
```
iurly: Perhaps you meant?
```suggestion
	download_from_artifactory)
```

probe_builder/builder/distro/debian.py
```diff
@@ -0,0 +1,112 @@
...
+            self._reparent_link(base_path, release, 'source')
+
+            makefile = os.path.join(kerneldir, 'Makefile')
+            makefile_orig = makefile + '.sysdig-orig'
+            target_in_container = target.replace(workspace.workspace, '/build/probe')
+            if not os.path.exists(makefile_orig):
+                orig = open(makefile).read()
+                open(makefile_orig, 'w').write(orig)
+                patched = orig.replace('/usr/src', os.path.join(target_in_container, 'usr/src'))
+                open(makefile, 'w').write(patched)
```
iurly: here we might want to use `with` for opening files or else (I guess) we're leaking file descriptors

iurly (changes requested):
First of all this is great, thanks!
Apart form a couple of nits, a few thoughts mainly due to my ignorance on the topic (and which should probably be addressed as a separate step/PR):
- I'm still struggling to understand the workflow and the implications. Perhaps having a dedicated presentation and/or a diagram might be helpful.
- was wondering whether we could move files around into a different folder hierarchy, so to better match the inner workings. So for instance Dockerfiles for building should probably live in a different directory than the top-level Dockerfile.
- why isn't the main entrypoint written in python as well? apart from historical reasons of course
- the whole volume mapping when using docker in docker is, to put it mildly, a nightmare. AFAICT it is due to the daemon architecture nature of docker. Perhaps for the actual builders (so inner containers) it might be worth exploring podman, which is daemonless?

gnosek:
> I'm still struggling to understand the workflow and the implications. Perhaps having a dedicated presentation and/or a diagram might be helpful.

I'll try to draw something. FWIW, it's basically unchanged from the old one, except that we pass richer metadata from the crawler to the builder.

> was wondering whether we could move files around into a different folder hierarchy, so to better match the inner workings. So for instance Dockerfiles for building should probably live in a different directory than the top-level Dockerfile.

I've been staring at this so long I've ran out of ideas multiple times ;) if you have any suggestions about changes to the layout, I'm all ears!

> why isn't the main entrypoint written in python as well? apart from historical reasons of course

No fundamental reason, really. I just wanted to avoid two separate CLIs in the same tool (which I'd have to do to have a decent new interface and keep compatibility with the old one). If you have strong feelings about it, I can try to get rid of the shell wrapper (there's not much left in it anyway).

> the whole volume mapping when using docker in docker is, to put it mildly, a nightmare. AFAICT it is due to the daemon architecture nature of docker. Perhaps for the actual builders (so inner containers) it might be worth exploring podman, which is daemonless?

There's one spot where we need to use docker :( we need --privileged to access the flatcar images. I was looking for pure userspace tools to copy files out of an ext4 filesystem and the closest I got was guestfish (which spawns an entire VM) and an extremely WIP tool that doesn't work. If we can work around that, I'd be extremely happy to rip out all docker references.

iurly (approved):
Way better than it used to be!
Let's yolo-merge it!

PR #19, opened 2021-10-20
[SSPROD-8709] Use repository metadata in kernel crawler, take 2

gnosek:
Summary of changes since the previous attempt:

+ Skip some more debian/ubuntu kernels
+ Exclude `kernel-modules` CentOS packages
+ Handle recursion manually for sqlite < 3.8.0
+ Python 3 fixes
+ Fedora kernels can use `kernel-modules` RPMs too
+ Install python3 in the builder container

### DIFF
--- a/kernel-crawler.py
+++ b/kernel-crawler.py
@@ -81,12 +81,14 @@ socket.getaddrinfo = getaddrinfo_ipv4only
 
 try:
     # noinspection PyCompatibility
-    from urllib2 import urlopen, unquote
+    from urllib2 import urlopen, unquote, HTTPError
 except ImportError:
     # noinspection PyCompatibility
     from urllib.request import urlopen
     # noinspection PyCompatibility
     from urllib.parse import unquote
+    # noinspection PyCompatibility
+    from urllib.error import HTTPError
 
 try:
     # noinspection PyCompatibility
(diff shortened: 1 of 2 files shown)


gnosek:
Verified this at least runs on the probe-builder box

deepskyblue86 (dismissed):
Looks good (I compared it to f1bb0e4) even if I have no clue about the query :)

kernel-crawler.py
```diff
@@ -624,22 +624,50 @@ def build_base_query(cls, version=''):
...
+                    if not new_ids:
+                        break
+                    pkgkeys.update(new_ids)
+                
+                placeholders = ', '.join(['?'] * len(pkgkeys))
+                query = '''SELECT location_href FROM packages WHERE pkgkey IN({})'''.format(placeholders)
+                cursor.execute(query, list(pkgkeys))
+                for row in cursor.fetchall():
+                    repo_db.append((ver, row[0]))
+            return repo_db
```
iurly: Is there any particular reason for having to maintain two different versions of the code which should be doing exactly the same thing?
I mean, is the recursive query significantly more efficient than the one performed by hand?
If not, I'd rather just always use the manually unfolded version (and perhaps keep the cleaner one commented as a future reference for when we'll want to deprecate older sqlite3 versions).
Just my two cents.

gnosek: I commented the recursive variant out. Maybe we'll be able to switch to it in the future

iurly:
@gnosek I guess this PR is now obsolete, can we close it?

PR #20, opened 2021-10-20
Add a Debian bullseye builder

gnosek:
This should ship with `gcc-10` already present, so we don't need to symlink it manually


deepskyblue86 (approved):

jcpittman144 (approved):

gnosek:
We already have one (from a different PR I suppose). Closing

PR #21, opened 2021-10-20
[SSPROD-9627] Enhance build-probe-binaries Debian logic to handle new linux-headers Makefile format

jcpittman144:
Fixes to allow build-probe-binaries -k CustomDebian to work on Debian 11:
- Linux headers for Debian 11 use a new format for the architecture-specific Makefile.
  Enhance the Makefile relocation patching logic to handle this new format as well as the old one.
- Fix build-probe-binaries Debian linux-kbuild finder logic to accomodate package location
  in the CustomDebian case.
- Fix Debian Dockerfiles to install libelf-dev package (like Ubuntu Dockerfiles)
- Add new Dockerfile.debian-gcc10.2, to handle the version of GCC available in Debian 11.

### DIFF
--- a/Dockerfile.debian-gcc10.2
+++ b/Dockerfile.debian-gcc10.2
@@ -1,4 +1,4 @@
-FROM debian:jessie
+FROM debian:bullseye
 
 RUN apt-get update && apt-get -y --no-install-recommends install \
 	cmake \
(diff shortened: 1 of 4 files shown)


gnosek (approved):
✅

PR #26, opened 2021-12-03
Skip eBPF probes for Fedora 35, fix Fedora 34 eBPF

gnosek:
###    [SMAGENT-3278] Don't install clang in Fedora 35 container

    The eBPF probe builds successfully with clang 13 but then fails
    the verifier test upon loading.

###    [SMAGENT-3278] Use specific builder for Fedora 34

    Fedora 34 and 35 have the same gcc version (at least at this point
    in time) but are otherwise incompatible:
    - fc34 can't build probes for fc35 kernels because fc35 needs a newer glibc
    - fc35 can't build eBPF probes (at all) because its clang is too new

    1. Try to parse the kernel version to look for distro-specific tags.
    If one is found, look for a matching dockerfile. If there is one, use it.
    Otherwise, fall back to gcc version based choice.

    2. Add a fc34 builder

[SMAGENT-3278]: https://sysdig.atlassian.net/browse/SMAGENT-3278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SMAGENT-3278]: https://sysdig.atlassian.net/browse/SMAGENT-3278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile.centos-gcc11.2
+++ b/Dockerfile.centos-gcc11.2
@@ -13,8 +13,6 @@ RUN yum -y install \
 	elfutils-devel \
 	findutils \
 	kmod \
-	clang \
-	llvm \
 	python-lxml && yum clean all
 
 ADD builder-entrypoint.sh /
--- a/Dockerfile.fc34
+++ b/Dockerfile.fc34
@@ -1,0 +1,22 @@
+FROM fedora:34
+
+RUN yum -y install \
+	wget \
+	git \
+	gcc \
+	gcc-c++ \
+	autoconf \
+	bison \
+	flex \
+	make \
+	cmake \
+	elfutils-devel \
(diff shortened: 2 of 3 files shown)


deepskyblue86 (approved):
Makes sense. :crossed_fingers:

jcpittman144 (commented):
Approach seems fine to me.

jcpittman144 (approved):

PR #29, opened 2022-01-17
Option to enforce probe download instead of building it

francesco-racciatti:
This add an option to modify the workflow of the sysdig-probe-loader script.

When the environment variable `SYSDIG_FORCE_DOWNLOAD_PROBE` is defined,
the script tries to download the appropriate kernel/bpf probe, instead of building it.

### DIFF
--- a/sysdig-probe-loader
+++ b/sysdig-probe-loader
@@ -33,6 +33,11 @@
 # 3) Fetch the appropriate sysdig-probe (kernel module or eBPF) for the
 #    currently running kernel, from download.sysdig.com
 #
+# Workflow variations:
+# A) If the environment variable SYSDIG_FORCE_DOWNLOAD_PROBE is defined,
+#    as the first and ONLY step, the script tries to download the appropriate sysdig-probe (kernel module or eBPF).
+#    The script won't try to build the probe.
+#
 #
 # --- ENVIRONMENT VARIABLES ---
 #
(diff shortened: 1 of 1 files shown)


gnosek:
I like it, but I'm afraid the indentation is getting (even more) excessive. Can we introduce some helper functions and call them conditionally instead of the large inline `if` bodies?

francesco-racciatti:
Yep, absolutely!
tks

iurly:
@gnosek I was also wondering if should try and have an overall variable to define the strategy in some self-explanatory way, instead of adding yet another cryptical boolean one which makes things even harder to decipher.
For instance, something like `SYSDIG_PROBE_STRATEGY`, `SYSDIG_PROBE_SOURCE`, `SYSDIG_PROBE_BEHAVIOR`.
Example values would be `"ebpf:build,download"` or `"kmod:download,none"`.
Of course we would have to retain compatibility if already-known variables are set, and infer the strategy from there.
cc: @balinaveen @alblasco

gnosek:
I'm not sure. I mean, I like the idea but we'll end up with a parser for these values written in shell and if you put these words in a single sentence and rearrange them a bit, they're going to spell trouble ;)

francesco-racciatti:
@gnosek 
In the meantime I refactored the whole script.

francesco-racciatti:
@iurly @gnosek
I refactored the script.

sysdig-probe-loader
```diff
@@ -557,24 +592,8 @@ load_bpf_probe() {
...
-
-		echo "* BPF probe located, it's now possible to start sysdig"
-
-		ln -sf "${HOME}/.sysdig/${BPF_PROBE_FILENAME}" "${HOME}/.sysdig/${BPF_PROBE_NAME}.o"
-		exit $?
-	else
-		echo "* Failure to find a BPF probe"
-		exit 1
-	fi
+	make_symlink_bfp_probe
```
gnosek: s/bfp/bpf/ :) (here and above)

gnosek (commented):
Apart from the one typo (bfp -> bpf) it looks good :)

With a change this large (basically rewriting the whole thing) multiple individual commits are nice but I think we've gone a bit too far in this case :)

For one thing, there are commits that are reverted immediately, like the load -> mount -> load rewording, and also there's a lot of comment/log changes, which I guess for the most part could be left alone

sysdig-probe-loader
```diff
@@ -48,7 +58,6 @@
 # Where:
 #   DOWNLOAD_URL_HOST = "download.sysdig.com" by default
 #   DOWNLOAD_REPOSITORY = "stable" by defualt
```
alblasco: typo: defualt

sysdig-probe-loader
francesco-racciatti: yes, I agree, I exaggerated a little ... 
I will pay more attention! :)

sysdig-probe-loader
francesco-racciatti: Thanks, fixed!

francesco-racciatti:
@alblasco @gnosek 
Thanks!  I fixed the typos.

gnosek (approved):
✅

PR #30, opened 2022-01-28
Option to enforce probe building instead of downloading it and preventing installing leftover probes

francesco-racciatti:
This PR adds an option to modify the workflow of the sysdig-probe-loader script.

When the environment variable `SYSDIG_FORCE_BUILD_PROBE` is defined,
the script tries to build the appropriate kernel/bpf probe, and makes no attempts to download it.

Moreover, this PR ensures that the script will not install leftover probes when using 
SYSDIG_FORCE_BUILD_PROBE/SYSDIG_FORCE_DOWNLOAD_PROBE.


gnosek (approved):
LGTM

PR #33, opened 2022-02-14
Fix deb crawler py3 compatibility

gnosek:


probe_builder/kernel_crawler/deb.py
```diff
@@ -247,4 +250,4 @@ def list_repos(self):
                 except requests.HTTPError:
                     pass
 
-        return sorted(repos.values())
+        return sorted(repos.values(), key=str)
```
deepskyblue86, line 253: oh my, is this still a thing? :facepalm:

deepskyblue86 (approved):

PR #34, opened 2022-02-24
Python builder road test

iurly:
This PR addresses several issues found while trying to put the new python builder into production.
Each individual commit message provides an in-depth description of the issue it fixes.

### DIFF
--- a/probe_builder/__init__.py
+++ b/probe_builder/__init__.py
@@ -61,7 +61,7 @@ class CrawlDistro(object):
     def get_kernels(self, workspace, _packages, download_config):
         # crawler will return a dictionary of {"release": ["http://url/for/package1.rpm", "http://url/for/package1.rpm"]}
         crawled_dict = self.distro_builder.crawl(workspace, self.distro_obj, self.crawler_distro, download_config)
-        # we flatten that dictionary into a single list, retaining ONLY package urls and discaring the release altogether
+        # we flatten that dictionary into a single list, retaining ONLY package urls and discarding the release altogether
         flattened_packages = [pkg for pkgs in crawled_dict.values() for pkg in pkgs]
         # then we batch that list as if it were a local distro
         batched_packages = self.distro_builder.batch_packages(flattened_packages)
--- a/probe_builder/__init__.py
+++ b/probe_builder/__init__.py
@@ -59,8 +59,13 @@ class CrawlDistro(object):
         self.crawler_distro = crawler_distro
 
     def get_kernels(self, workspace, _packages, download_config):
-        return self.distro_builder.crawl(workspace, self.distro_obj, self.crawler_distro, download_config)
-
+        # crawler will return a dictionary of {"release": ["http://url/for/package1.rpm", "http://url/for/package1.rpm"]}
+        crawled_dict = self.distro_builder.crawl(workspace, self.distro_obj, self.crawler_distro, download_config)
+        # we flatten that dictionary into a single list, retaining ONLY package urls and discaring the release altogether
+        flattened_packages = [pkg for pkgs in crawled_dict.values() for pkg in pkgs]
+        # then we batch that list as if it were a local distro
+        batched_packages = self.distro_builder.batch_packages(flattened_packages)
+        return batched_packages
 
 class LocalDistro(object):
 
--- a/probe_builder/builder/distro/debian.py
+++ b/probe_builder/builder/distro/debian.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import re
+import traceback
 
 import click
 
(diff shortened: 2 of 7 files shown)


probe_builder/__init__.py
```diff
@@ -59,8 +59,13 @@ def __init__(self, distro, builder_distro, crawler_distro):
         self.crawler_distro = crawler_distro
 
     def get_kernels(self, workspace, _packages, download_config):
-        return self.distro_builder.crawl(workspace, self.distro_obj, self.crawler_distro, download_config)
-
+        # crawler will return a dictionary of {"release": ["http://url/for/package1.rpm", "http://url/for/package1.rpm"]}
+        crawled_dict = self.distro_builder.crawl(workspace, self.distro_obj, self.crawler_distro, download_config)
+        # we flatten that dictionary into a single list, retaining ONLY package urls and discaring the release altogether
```
gnosek:
```suggestion
        # we flatten that dictionary into a single list, retaining ONLY package urls and discarding the release altogether
```

probe_builder/builder/distro/debian.py
```diff
@@ -95,21 +104,31 @@ def batch_packages(self, kernel_files):
...
             version = m.group('version')
-            arch = m.group('arch')
+            vararch = m.group('vararch')
 
-            if arch == 'common':
+            if 'common' in vararch:
+                #
+                # linux-headers-5.16.0-1-|common|_5.16.7-2_all.deb
+                # linux-headers-5.16.0-1-|common-rt|_5.16.7-2_all.deb
+                #
```
gnosek, line 113: I feel this should be based on the package architecture, not the name (i.e. `all` vs `amd64` right before the final dot)

probe_builder/builder/distro/debian.py
```diff
@@ -95,21 +104,31 @@ def batch_packages(self, kernel_files):
...
+        for version, per_vararch in arch_packages.items():
+            for vararch, packages in per_vararch.items():
                 packages.extend(common_packages.get(version, []))
                 major, minor, _ = version.split('.', 2)
                 major_version = '{}.{}'.format(major, minor)
                 kbuild_pkg = kbuild_packages.get(major_version)
                 if kbuild_pkg:
                     packages.append(kbuild_pkg)
-                kernels['{}-{}'.format(version, arch)] = packages
+                kernels['{}:{}'.format(version, vararch)] = packages
```
gnosek, line 132: Any chance we could use a tuple `(version, vararch)` as the dict key? It would make things a bit less stringly typed

probe_builder/kernel_crawler/download.py
```diff
@@ -61,7 +62,10 @@ def dl():
         while True:
             url = q.get()
             output_file = os.path.join(output_dir, os.path.basename(url))
-            download_file(url, output_file, download_config)
+            try:
+                download_file(url, output_file, download_config)
+            except requests.exceptions.RequestException:
+                traceback.print_exc()
```
gnosek, line 68: do we want to delete the file here too?

probe_builder/kernel_crawler/download.py
```diff
@@ -73,6 +77,9 @@ def dl():
         q.put(batch_url)
 
     q.join()
+    # TODO shouldn't we also wait for all threads to terminate,
+    # like it's done on the examples of python 3.7 (but NOT on 2.7/3.8/3.9)?
+    # see https://docs.python.org/3.7/library/queue.html
```
gnosek, line 82: I think right now they do *not* terminate and just stay hanging in the background (they're spinning in a `while True` after all). IIRC I couldn't find an easy way to tell them "there's nothing more happening" but now that I think about it, we might be able to e.g. push `None` to the queue as many times as there are threads and handle it inside `dl`.

gnosek (commented):
Some minor comments but overall, it's great 🤩

probe_builder/builder/distro/debian.py
iurly: I agree.
My initial take on this was that we would be having different "versions" (i.e. `5.16.0-1` and `5.16.0-1-rt` so that `common-rt` would belong to `5.16.0-1-rt` only and I would need to drop the `common` prefix.
But then I realized they're just different pseudo-architectures of the same version and they all belong together, so that part of the code is gone and this doesn't make much sense anymore.

probe_builder/builder/distro/debian.py
iurly: I thought about that too, but I would save it for a subsequent refactoring.
I'm also not sure why we need different approaches for ubuntu and debian.

probe_builder/kernel_crawler/download.py
iurly: Not sure. There's also some retrying/resume logic if you get interrupted at the middle of the download.
Perhaps we might want to delete it if the download never started to begin with (e.g. 404) and you're left with a zero-sized file?

probe_builder/kernel_crawler/download.py
iurly: Exactly, that's what the example does.

deepskyblue86 (dismissed):
LGTM (just don't forget to fix the typo).
Moreover, this is quite urgent.

deepskyblue86 (approved):

PR #36, opened 2022-03-03
fix ubuntu missing kernels, take2

iurly:
- builder: allow multiple target directories for the same release
- ubuntu: treat each package version as a separate release


gnosek (approved):
LGTM

PR #46, opened 2022-03-21
Optimize builders

iurly:
This PR adds a few optimizations to slightly speed up the build step process:
- only build the builders once
- do not try to build bpf probes if the builder does not explicitly support it
- add a builder for debian 12 bookworm


probe_builder/builder/builder_image.py
```diff
@@ -7,11 +7,36 @@
 
 logger = logging.getLogger(__name__)
 
+# TODO here perhaps we might want to make a class?
```
deepskyblue86, line 10: pending TODO

probe_builder/builder/builder_image.py
```diff
@@ -7,11 +7,36 @@
...
+    obj = True
+    # ... but if we make this a factory @staticmethod of a proper class, we'll have to do this:
+    #obj = cls()
+
     image_name = '{}sysdig-probe-builder:{}'.format(workspace.image_prefix, dockerfile_tag)
-    docker.build(image_name, dockerfile, workspace.builder_source)
+    if workspace.image_prefix:
+        # image_prefix essentially means: get this from a docker repo
+        pass
+        # TODO here we should have something like
```
deepskyblue86: and why aren't we doing it?

probe_builder/builder/choose_builder.py
```diff
@@ -113,17 +115,23 @@ def choose_gcc_dockerfile(builder_source, builder_distro, kernel_dir):
     #   10.0
     # and now we properly realize that gcc 10 is newer than 9.2, not older than 4.4
     prefix = 'Dockerfile.{}-gcc'.format(builder_distro)
-    dockerfile_versions = [Version(f[len(prefix):]) for f in os.listdir(builder_source) if f.startswith(prefix)]
-    dockerfile_versions.sort()
+    regex = re.compile('^' + re.escape(prefix) + '(?P<gccver>[0-9]+\.[0-9]+)(?P<bpf>(\-bpf)?)$')
```
deepskyblue86, line 126: update the comment above, maybe also mentioning why we have a `bpf` group that we don't use

probe_builder/builder/choose_builder.py
```diff
@@ -113,17 +115,23 @@ def choose_gcc_dockerfile(builder_source, builder_distro, kernel_dir):
...
+    dockerfile_versions = [
+        (Version(regex.match(f).group('gccver')), f)
+        for f in os.listdir(builder_source)
+        if regex.match(f)
```
deepskyblue86: I'd like it better in a single line as it was before

probe_builder/builder/distro/base_builder.py
```diff
@@ -88,8 +88,11 @@ def build_kernel(self, workspace, probe, builder_distro, release, target):
         kernel_dir = self.get_kernel_dir(workspace, release, target)
         dockerfile, dockerfile_tag = choose_builder.choose_dockerfile(workspace.builder_source, builder_distro,
                                                                       kernel_dir)
-        if not workspace.image_prefix:
-            builder_image.build(workspace, dockerfile, dockerfile_tag)
+        # let build() figure out if it actually needs to build or pull anything
+        builder_image.build(workspace, dockerfile, dockerfile_tag)
```
deepskyblue86, line 92: Maybe have a different name, if `build` only eventually builds? IMHO the comment makes sense only by looking at the diff...

probe_builder/builder/distro/base_builder.py
```diff
@@ -88,8 +88,11 @@ def build_kernel(self, workspace, probe, builder_distro, release, target):
...
+
+        if not dockerfile.endswith('-bpf'):
+            ebpf_skip_reason = "Dockerfile {} does not end with -bpf".format(dockerfile)
```
deepskyblue86: I'd prefer a more meaningful message, since the -bpf suffix is just a convention we made up, it would have been on a different folder or something else.
"eBPF not supported by builder {}"?
"Builder {} does not support eBPF"?

deepskyblue86 (commented):
I don't have enough context, I'll give another look at this later

probe_builder/builder/builder_image.py
iurly: my mistake, I don't think that's the intended use case

probe_builder/builder/builder_image.py
iurly: yes, in the future

probe_builder/builder/choose_builder.py
iurly: I changed the logic so to actually use that information

probe_builder/builder/distro/base_builder.py
iurly: done

probe_builder/builder/builder_image.py
deepskyblue86: I don't really like the idea of committing future purposes (either the comment either the commented out code)

deepskyblue86 (approved):

PR #47, opened 2022-03-22
Parallel builders

iurly:


probe_builder/__init__.py
```diff
@@ -96,7 +97,6 @@ def get_kernels(self, _workspace, packages, _download_config, filter):
 def cli(debug):
     init_logging(debug)
 
-
```
gnosek: If you touch this file again, maybe remove this change :)

probe_builder/__init__.py
```diff
@@ -124,8 +124,16 @@ def build(builder_image_prefix,
...
-        distro_builder.build_kernel(workspace, probe, distro.builder_distro, release, target)
+
+    parallelism = len(os.sched_getaffinity(0))
+    with ThreadPoolExecutor(max_workers=parallelism) as executor:
+        kernels_futures = []
+        for release, target in kernel_dirs:
+            future = executor.submit(distro_builder.build_kernel, workspace, probe, distro.builder_distro, release, target)
+            kernels_futures.append((release, future))
+
+    for release, future in kernels_futures:
```
gnosek, line 136: I know ~nothing about python futures but I assume this waits for the futures to be resolved, right?

probe_builder/__init__.py
```diff
@@ -124,8 +124,16 @@ def build(builder_image_prefix,
...
+        print("Release: {}, Result: {}".format(release, future.result()))
```
gnosek, line 137: Do we want to use the logger here?

probe_builder/builder/choose_builder.py
```diff
@@ -146,6 +148,7 @@ def choose_gcc_dockerfile(builder_source, builder_distro, kernel_dir):
     dockerfile, _, support_bpf = dockerfile_version
     # get a tag we'll use for builders
     tag = dockerfile.replace('Dockerfile.', '')
+    logger.debug('elected dockerfile: {}'.format(dockerfile))
```
gnosek: chosen? selected?

probe_builder/builder/distro/base_builder.py
```diff
@@ -88,14 +91,21 @@ def build_kernel(self, workspace, probe, builder_distro, release, target):
...
         dockerfile, dockerfile_tag, support_bpf = choose_builder.choose_dockerfile(workspace.builder_source, builder_distro,
                                                                       kernel_dir)
+
+        ts0 = time.time()
         # let build() figure out if it actually needs to build or pull anything
         builder_image.build(workspace, dockerfile, dockerfile_tag)
+        ts1 = time.time()
+
+        logger.info("Docker building of {} took {:.2f}s".format(dockerfile, ts1-ts0))
+        took = ts1 - ts0
```
gnosek:
```suggestion
        took = ts1 - ts0
        logger.info("Docker building of {} took {:.2f}s".format(dockerfile, took))
```

probe_builder/builder/distro/base_builder.py
```diff
@@ -54,15 +55,17 @@ def build_kernel_impl(cls, config_hash, container_name, image_name, kernel_dir,
             logger.info('Skipping build of {} probe {}-{}: {}'.format(label, release, config_hash, skip_reason))
             return
 
-        docker.rm(container_name)
+        #docker.rm(container_name)
```
gnosek, line 58: Maybe just delete it?

probe_builder/builder/distro/base_builder.py
```diff
@@ -88,14 +91,21 @@ def build_kernel(self, workspace, probe, builder_distro, release, target):
...
+        logger.info("Docker building of {} took {:.2f}s".format(dockerfile, ts1-ts0))
+        took = ts1 - ts0
 
         if not support_bpf:
             ebpf_skip_reason = "Builder {} does not support eBPF".format(dockerfile)
 
         image_name = '{}sysdig-probe-builder:{}'.format(workspace.image_prefix, dockerfile_tag)
-        container_name = 'sysdig-probe-builder-{}'.format(dockerfile_tag)
+        #container_name = 'sysdig-probe-builder-{}'.format(dockerfile_tag)
+        container_name = ''
```
gnosek, line 108: How worried are we about leftover containers after e.g. an interrupted build?

probe_builder/__init__.py
```diff
@@ -124,8 +124,16 @@ def build(builder_image_prefix,
 
     kernels = distro_obj.get_kernels(workspace, package, download_config, filter)
     kernel_dirs = distro_builder.unpack_kernels(workspace, distro.distro, kernels)
-    for release, target in kernel_dirs:
-        distro_builder.build_kernel(workspace, probe, distro.builder_distro, release, target)
+
+    parallelism = len(os.sched_getaffinity(0))
```
gnosek: Maybe make it configurable, so (worst case) we can disable parallelism if we need to?

gnosek (dismissed):
Some minor nits but it's beautiful 🤩

probe_builder/__init__.py
iurly: good question, I didn't give it too much thought. I guess I went straight for a print() because that's essentially a final report, so there would be no added value in having a timestamp in each line.
At the end of the day we might even want to produce some structured artifact to be archived, such as an XML or HTML file. What do you think?

probe_builder/builder/distro/base_builder.py
iurly: thanks! I actually also caught a serious error while looking this up!

probe_builder/builder/distro/base_builder.py
iurly: good point. I guess the real threat is when you forcibly kill the main builder, in which case this won't really help I'm afraid (except at the next attempt).
We're ultimately using `docker --rm` which should clean up after itself, but you're right in that the following scenario will leave us with zombies:
- main builder spawned
- <max_jobs> probe builders spawned
- one or more of the above builders get stuck for whatever reason
- main builder gets killed by timeout or impatient user

at this point our best bet would be to have some external cleanup mechanism (either upon start or as a post step), based on either (a) the image name or (b) a prefix-based container name (e.g. container-name + random number).
I'd go for option (a), don't see much added value in (b).

probe_builder/__init__.py
iurly: This way it *sort of is* configurable. You just need to call the builder with the right value for `taskset`. :trollface: 
OK we already have a parameter for download parallelism, shall we add another one for build parallelism?
And yet another one for unpacker parallelism?
We're just running out of one-letter parameters but it all makes sense I guess.

probe_builder/builder/distro/base_builder.py
iurly: Changes accepted with the latest rework (from ~1 month ago)

builder-entrypoint.sh
```diff
@@ -60,11 +60,17 @@ build_bpf() {
...
+		mkdir -p /build/sysdig
+		cd /build/sysdig
+		call_cmake /code/sysdig-rw
+		make -C /code/sysdig-rw/driver/bpf clean all
+		cp /code/sysdig-rw/driver/bpf/probe.o $OUTPUT/$PROBE_NAME-bpf-$PROBE_VERSION-$ARCH-$KERNEL_RELEASE-$HASH.o
 	fi
 }
 
+rm -rf /code/sysdig-rw
+cp -rf /code/sysdig-ro /code/sysdig-rw
```
deepskyblue86, line 72: :open_mouth: :heart_decoration:

probe_builder/__init__.py
deepskyblue86: @gnosek `result()` is the thing https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result

probe_builder/__init__.py
gnosek: Still, I'd rather have an option to run things sequentially if we run into issues, instead of reverting the PR.

probe_builder/builder/choose_builder.py
gnosek: bump

gnosek (dismissed):
Nothing major, but I'd still prefer an option to do sequential builds if we run into issues

probe_builder/__init__.py
iurly: done

probe_builder/__init__.py
iurly: done, I chose `-j`

probe_builder/builder/choose_builder.py
iurly: ok done u1

iurly (commented):
@gnosek stage's all yours now!

gnosek (approved):

PR #48, opened 2022-03-23
rework parallel downloads

iurly:
The current implementation fails when issuing parallel downloads
in that there might be multiple URLs for the same target file.
Rework the whole scheduling so that URLs are grouped by target file,
and multiple URLs for the same file are only attempted sequentially.

While at it, use futures as opposed to explicit threading.


gnosek (approved):
❤️

PR #50, opened 2022-04-11
handle new driver's parameter naming

iurly:
the new driver with semantic versioning does not currently
compile because cmake parameter names have been renamed.
So in order to have something working with the existing naming
scheme, use the new parameter names with the old semantics.
Which I have to admin, feels like defeating the whole purpose
of the renaming. But at least we'd get probes for the next release.


builder-entrypoint.sh
```diff
@@ -14,6 +14,16 @@ set -euo pipefail
...
 
+call_cmake() {
+	SRC_DIR=$1
+	if [[ -f ${SRC_DIR}/driver/API_VERSION ]]; then
+		# New naming
+		cmake -DCMAKE_BUILD_TYPE=Release -DDRIVER_NAME=$PROBE_NAME -DDRIVER_VERSION=$PROBE_VERSION -DDRIVER_DEVICE_NAME=$PROBE_DEVICE_NAME -DCREATE_TEST_TARGETS=OFF ${SRC_DIR}
+	else
+		cmake -DCMAKE_BUILD_TYPE=Release -DPROBE_NAME=$PROBE_NAME -DPROBE_VERSION=$PROBE_VERSION -DPROBE_DEVICE_NAME=$PROBE_DEVICE_NAME -DCREATE_TEST_TARGETS=OFF ${SRC_DIR}
+	fi
+}
```
deepskyblue86:
```suggestion
call_cmake() {
	SRC_DIR=$1
	PROBE_NAME_PARAM=PROBE_NAME
	PROBE_VERSION_PARAM=PROBE_VERSION
	PROBE_DEVICE_NAME_PARAM=PROBE_DEVICE_NAME
	if [[ -f ${SRC_DIR}/driver/API_VERSION ]]; then
		# New param naming
		PROBE_NAME_PARAM=DRIVER_NAME
		PROBE_VERSION_PARAM=DRIVER_VERSION
		PROBE_DEVICE_NAME_PARAM=DRIVER_DEVICE_NAME
	fi
	cmake -DCMAKE_BUILD_TYPE=Release -D${PROBE_NAME_PARAM}=$PROBE_NAME -D${PROBE_VERSION_PARAM}=$PROBE_VERSION -D${PROBE_DEVICE_NAME_PARAM}=$PROBE_DEVICE_NAME -DCREATE_TEST_TARGETS=OFF ${SRC_DIR}
}
```

builder-entrypoint.sh
```diff
@@ -25,7 +35,7 @@ build_kmod() {
 	mkdir -p /build/sysdig
 	cd /build/sysdig
 
-	cmake -DCMAKE_BUILD_TYPE=Release -DPROBE_NAME=$PROBE_NAME -DPROBE_VERSION=$PROBE_VERSION -DPROBE_DEVICE_NAME=$PROBE_DEVICE_NAME -DCREATE_TEST_TARGETS=OFF /build/probe/sysdig
```
deepskyblue86, line 28: I see the same line in `builder-entrypoint-coreos.sh`, is that deprecated?

deepskyblue86 (changes requested):

builder-entrypoint.sh
deepskyblue86: Looks not used right now, let's move on.

deepskyblue86 (approved):

PR #51, opened 2022-04-22
Dedicated AmazonLinux2 builder

gnosek:
AmazonLinux2 ships with gcc 7.3, which makes us choose
the builder image based on Fedora 27. That comes with clang 5,
which does build the eBPF probe successully but the probe
fails verification, e.g. with kernel 4.14.275-207.503.amzn2.

Create a dedicated AmazonLinux2 builder, with clang 7 (as it's
more battle tested than clang 11 which comes in the default
clang package on AmazonLinux2).

### DIFF
--- a/Dockerfile.amzn2-gcc7.3-bpf
+++ b/Dockerfile.amzn2-gcc7.3-bpf
@@ -1,0 +1,24 @@
+FROM amazonlinux:2
+
+RUN yum -y install \
+	wget \
+	git \
+	gcc \
+	gcc-c++ \
+	autoconf \
+	bison \
+	flex \
+	make \
+	cmake \
+	elfutils-devel \
+	findutils \
+	kmod \
+	clang7.0 \
+	llvm7.0 \
+	python-lxml && yum clean all
+
+ADD builder-entrypoint.sh /
+ENV CLANG clang-7
+ENV LLC llc-7.0
+WORKDIR /build/probe
+ENTRYPOINT [ "/builder-entrypoint.sh" ]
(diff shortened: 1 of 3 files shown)


Dockerfile.amzn2-gcc7.3-bpf
```diff
@@ -0,0 +1,24 @@
...
+	cmake \
+	elfutils-devel \
+	findutils \
+	kmod \
+	clang7.0 \
+	llvm7.0 \
+	python-lxml && yum clean all
+
+ADD builder-entrypoint.sh /
+ENV CLANG clang-7
```
iurly, line 23:
```suggestion
# Enforce usage of clang 7, as it's more battle tested than clang 11
# which comes in the default clang package on AmazonLinux2
ENV CLANG clang-7
```

iurly (dismissed):
LGTM! I'd just add a comment in the code for posterity (taken from the commit message)

iurly (approved):

Dockerfile.amzn2-gcc7.3-bpf
```diff
@@ -0,0 +1,26 @@
...
+	kmod \
+	clang7.0 \
+	llvm7.0 \
+	python-lxml && yum clean all
+
+ADD builder-entrypoint.sh /
+# Enforce usage of clang 7, as it's more battle tested than clang 11
+# which comes in the default clang package on AmazonLinux2
+ENV CLANG clang-7
+ENV LLC llc-7.0
```
deepskyblue86, line 24: I stumbled upon this yesterday. It makes me crazy that clang is `clang-7` but llc is `llc-7.0`!

deepskyblue86 (approved):

PR #53, opened 2022-04-22
PhotonOS generic kernel support

deepskyblue86:
Rework of #3. I left that one in place because it might have some useful info for the other kernel variants.

### DIFF
--- a/probe_builder/builder/choose_builder.py
+++ b/probe_builder/builder/choose_builder.py
@@ -96,7 +96,7 @@ def choose_gcc_dockerfile(builder_source, builder_distro, kernel_dir):
     kernel_gcc = get_kernel_gcc_version(kernel_dir)
     # We don't really care about the compiler patch levels, only the major/minor version
     kernel_gcc = Version(kernel_gcc)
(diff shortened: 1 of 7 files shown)


probe_builder/builder/choose_builder.py
```diff
@@ -96,6 +96,7 @@ def choose_gcc_dockerfile(builder_source, builder_distro, kernel_dir):
     kernel_gcc = get_kernel_gcc_version(kernel_dir)
     # We don't really care about the compiler patch levels, only the major/minor version
     kernel_gcc = Version(kernel_gcc)
+    logger.debug('found gcc version {}'.format(kernel_gcc))
```
iurly:
```suggestion
    logger.debug('found kernel gcc version {}'.format(kernel_gcc))
```

iurly (dismissed):

iurly (approved):
So much better now!

deepskyblue86:
Some stats here:
```sh
$ du -hs *
19G     build
654M    output
5.6G    photonos
```
At the time of writing, there are 161 kernels, hence 322 probes.

PR #54, opened 2022-06-01
add AmazonLinux 2022.0.20220419 2022.0.20220504

iurly:


gnosek (approved):
👍 

How do we get the new releases? Any chance we can script that?

iurly:
> How do we get the new releases? Any chance we can script that?

For the latest one I used:
```
docker run -it --rm amazonlinux:2022 python3 -c 'import dnf, json; db = dnf.dnf.Base(); print(json.dumps(db.conf.substitutions, indent=2))'
```

For the previous one, I used 
```
skopeo inspect docker://amazonlinux
```
which yielded something like:
```
....
        "2018.03.0.20220503.0-with-sources",
        "2022",
        "2022-with-sources",
        "2022.0.20220202.0",
        "2022.0.20220202.0-with-sources",
        "2022.0.20220308.1",
        "2022.0.20220308.1-with-sources",
        "2022.0.20220315.0",
        "2022.0.20220315.0-with-sources",
        "2022.0.20220419.0",
        "2022.0.20220419.0-with-sources",
        "2022.0.20220504.1",
        "2022.0.20220504.1-with-sources",
        "devel",
        "devel-with-sources",
        "latest",
        "with-sources"
....
```
and then I picked the image tag which looked more like the one we were missing, and re-ran the one-liner above, replacing `2022` -> `2022.0.20220419.0`

I can't think of a simpler process though.
So yes, all the above could be (more-or-less) easily scripted, though I'd be reluctant to doing it at every crawl, especially because it would introduce a dependency to `skopeo` (though we might want to use the docker registry API directly instead).
Perhaps some sort of depend-a-bot kind of thing instead?

PR #55, opened 2022-06-15
Amazon Linux 2022 dynamic release discovery

iurly:
After reading official documentation and a bit of tweaking
on an AmazonLinux2022 container image:

docker run -it --rm amazonlinux:2022
dnf check-release-update
dnf install 'dnf-command(check-release-update)'
cat ./usr/lib/python3.9/site-packages/dnf-plugins/release_notification.py

I was able to figure out that an XML file with the list of releases
could be easily obtained from a releasemd.xml.

See also https://github.com/aws/amazon-ecs-ami/blob/main/generate-release-vars.sh

So get the list of releases directly from there.

Notice how the repository of the latest release provides all kernel
packages (even the ones shipped on previous releases), so there's no
need to check ALL existing repositories -- we can use just the last one.

### DIFF
--- a/probe_builder/kernel_crawler/amazonlinux.py
+++ b/probe_builder/kernel_crawler/amazonlinux.py
@@ -5,6 +5,8 @@ import click
 
 from . import repo
 from . import rpm
+from lxml import etree, html
+import requests
 from probe_builder.kernel_crawler.download import get_url
 from probe_builder.py23 import make_string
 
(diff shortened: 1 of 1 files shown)


probe_builder/kernel_crawler/amazonlinux.py
```diff
@@ -63,23 +65,54 @@ class AmazonLinux2022Mirror(repo.Distro):
...
+        releasemd_url = "{}/{}".format(self.AL2022_BASE_URL, "core/releasemd.xml")
+        releasemd_xml = get_url(releasemd_url)
+        e = etree.fromstring(releasemd_xml)
+        # see https://github.com/aws/amazon-ecs-ami/blob/main/generate-release-vars.sh
+        # NOTE: the latest release alone seems to cumulatively provide kernel packages
+        # for all previous releases, too. Hence the last() bit.
+        expr = "//root/releases/release[last()]/@version"
+
+        # However, if you ever need to list ALL releases, use the following expression instead
+        expr = "//root/releases/release/@version"
```
francesco-racciatti, line 95: `expr` reassigned

francesco-racciatti, line 95: one of those `expr` should be commented out

probe_builder/kernel_crawler/amazonlinux.py
```diff
@@ -63,23 +65,54 @@ class AmazonLinux2022Mirror(repo.Distro):
...
-                repo_urls.add(get_al_repo("https://al2022-repos-us-east-1-9761ab97.s3.dualstack.us-east-1.amazonaws.com/core/mirrors/", r + '/x86_64'))
+                try:
+                    print("Adding repo {}".format(r), flush=True)
+                    repo_url = get_al_repo(
+                        "{}/{}".format(self.AL2022_BASE_URL, "core/mirrors/"),
+                        r + '/x86_64'
+                    )
+                    repo_urls.add(repo_url)
+                except requests.exceptions.HTTPError:
+                    print("WARNING: Could not get data for AmazonLinux2022 release: {}".format(r), flush=True)
```
francesco-racciatti: Could it make sense to log the exception message here?

```suggestion
                except requests.exceptions.HTTPError as err:
                    print("WARNING: Could not get data for AmazonLinux2022 release: {0}. Got error: {1}".format(r, err), flush=True)
```

probe_builder/kernel_crawler/amazonlinux.py
```diff
@@ -63,23 +65,54 @@ class AmazonLinux2022Mirror(repo.Distro):
...
+    # cat /etc/yum.repos.d/amazonlinux.repo
+    # https://al2022-repos-$awsregion-9761ab97.s3.dualstack.$awsregion.$awsdomain/core/mirrors/$releasever/$basearch/mirror.list
+    AL2022_BASE_URL = "https://al2022-repos-us-east-1-9761ab97.s3.dualstack.us-east-1.amazonaws.com"
 
     def __init__(self):
         super(AmazonLinux2022Mirror, self).__init__([])
 
     def list_repos(self):
+        # List of all available releases
+        releasemd_url = "{}/{}".format(self.AL2022_BASE_URL, "core/releasemd.xml")
```
alblasco, line 86: Just a question. I am used to use f-strings. I don't know if the python version or python compatibility prevents you from using f-strings

alblasco (dismissed):
Another comment.
Has any sense unit test on all new python code?

probe_builder/kernel_crawler/amazonlinux.py
iurly: Oh it's just a matter of personal preferences. 
I personally don't like f-strings as the use of identifiers buried in a string is somehow a violation of the old-fashioned semantics, at least IMHO.

probe_builder/kernel_crawler/amazonlinux.py
iurly: done

iurly:
> Another comment. Has any sense unit test on all new python code?

This code has currently no testing infrastructure whatsoever.
It would be indeed very useful, not only for testing our own code, but also for validating the structure of the data we're getting from third-party repositories, to make sure it hasn't changed under the hood.

francesco-racciatti (approved):

deepskyblue86 (approved):

PR #63, opened 2022-12-05
sysdig-probe-loader: work around Fedora 35+ SELinux restrictions

iurly:
Fedora 35+ distros on some cloud variants like AWS will carry some default SELinux policy preventing it from loading a .ko from a file descriptor through finit_module() (which is what insmod does).
Compressing it with .xz forces insmod to decompress it into memory first, and load it through init_module() instead, circumventing the issue.

### DIFF
--- a/sysdig-probe-loader
+++ b/sysdig-probe-loader
@@ -246,10 +246,24 @@ get_variable_override_kernel_probe() {
 #
 load_precompiled_kernel_probe() {
 	echo "* Loading module"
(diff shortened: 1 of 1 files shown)


francesco-racciatti (approved):

francesco-racciatti:
> circumventing the issue

circumventing the ~issue~ SELinux policy 😈

ironashram (approved):

gnosek (approved):

PR #68, opened 2023-01-24
add support for distro release filtering

iurly:
add an extra parameter to add the possibility to filter
packages for a given distro release.
The main use case is ubuntu, whose list of packages
(considering all releases) is simply humongous.

This way we can filter by adding -R "impish".

### DIFF
--- a/probe_builder/kernel_crawler/__init__.py
+++ b/probe_builder/kernel_crawler/__init__.py
@@ -32,10 +32,6 @@ DISTROS = {
     'Flatcar': FlatcarMirror,
 }
 
-
-def crawl_kernels(distro, distro_filter='', kernel_filter=''):
-    dist = DISTROS[distro]()
-    dist.set_distro_filter(distro_filter)
-    pt = dist.get_package_tree(kernel_filter)
-    dist.set_distro_filter('')
-    return pt
+def crawl_kernels(distro, crawler_filter):
+    dist = DISTROS[distro]
+    return dist().get_package_tree(crawler_filter)
--- a/probe_builder/kernel_crawler/__init__.py
+++ b/probe_builder/kernel_crawler/__init__.py
@@ -33,7 +33,9 @@ DISTROS = {
 }
 
 
-def crawl_kernels(distro, version=''):
-    dist = DISTROS[distro]
-
-    return dist().get_package_tree(version)
+def crawl_kernels(distro, distro_filter='', kernel_filter=''):
+    dist = DISTROS[distro]()
+    dist.set_distro_filter(distro_filter)
+    pt = dist.get_package_tree(kernel_filter)
+    dist.set_distro_filter('')
+    return pt
--- a/probe_builder/kernel_crawler/repo.py
+++ b/probe_builder/kernel_crawler/repo.py
@@ -4,9 +4,15 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 import click
 import sys
 
+class CrawlerFilter(object):
+    def __init__(self, distro_filter='', kernel_filter=''):
+        self.distro_filter = distro_filter
+        self.kernel_filter = kernel_filter
+
+EMPTY_FILTER=CrawlerFilter()
 
 class Repository(object):
-    def get_package_tree(self, kernel_filter=''):
+    def get_package_tree(self, crawler_filter):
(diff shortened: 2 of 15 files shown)


probe_builder/kernel_crawler/__init__.py
```diff
@@ -32,7 +32,9 @@
...
 
-def crawl_kernels(distro, version=''):
-    dist = DISTROS[distro]
-
-    return dist().get_package_tree(version)
+def crawl_kernels(distro, distro_filter='', kernel_filter=''):
+    dist = DISTROS[distro]()
+    dist.set_distro_filter(distro_filter)
+    pt = dist.get_package_tree(kernel_filter)
+    dist.set_distro_filter('')
```
gnosek: Do we need this line? AIUI, we're creating a new instance and `set_distro_filter` sets only an instance field so it should be safe to just let `dist` go out of scope without resetting it.

If I'm wrong, I'm happy to be enlightened :)

gnosek (commented):
I'm not a fan of setting `distro_filter` as an instance field, for two reasons:
* it's harder to reason about the lifetime of the setting when we set/reset it
* it introduces an asymmetry with the kernel filter, which is passed around as a parameter

Other than that, I love it!

iurly:
> I'm not a fan of setting `distro_filter` as an instance field, for two reasons:
> 
>     * it's harder to reason about the lifetime of the setting when we set/reset it
> 
>     * it introduces an asymmetry with the kernel filter, which is passed around as a parameter
> 
> 
> Other than that, I love it!

The first implementation was indeed a parameter just like the kernel filter.
The reason I didn't like it was that it changed the method signature for lots of sub classes, and it was then easy to forget to update some... especially the ones that came in through some pending PRs.
The instance field was the first thing that came to mind which didn't require to change every single distro.
Shall we make a generic class like a `CrawlerFilter` (carrying information about both distro & kernel filter, plus whatever we'll want to come up with in the future) instead?

probe_builder/kernel_crawler/__init__.py
iurly: Oh yes, you're right, that's not strictly needed. I just put it there for clarity.

gnosek:
> Shall we make a generic class like a `CrawlerFilter` (carrying information about both distro & kernel filter, plus whatever we'll want to come up with in the future) instead?

Yes please!

probe_builder/kernel_crawler/__init__.py
gnosek: For me personally, it decreases the clarity a little since it seems to suggest we're holding onto the `dist` object longer than we are

deepskyblue86: I'll argue on the clarity here :sweat_smile: 
This extra like makes me think it's needed while it's not, so the effect is more :thinking:  to me

probe_builder/kernel_crawler/repo.py
```diff
@@ -19,27 +19,39 @@ def to_s(s):
...
 
+
     def list_repos(self):
         repos = []
         with click.progressbar(
                 self.mirrors, label='Checking repositories', file=sys.stderr, item_show_func=to_s) as mirrors:
             for mirror in mirrors:
+                mirror.set_distro_filter(self.distro_filter)
                 repos.extend(mirror.list_repos())
+                mirror.set_distro_filter('')
```
deepskyblue86: do we need to reset it here?

deepskyblue86 (dismissed):

probe_builder/kernel_crawler/repo.py
```diff
@@ -3,9 +3,15 @@
 import click
 import sys
 
+class CrawlerFilter(object):
+    def __init__(self, distro_filter='', kernel_filter=''):
+        self.distro_filter = distro_filter
+        self.kernel_filter = kernel_filter
+
+EMPTY_FILTER=CrawlerFilter()
```
gnosek: Since we're using `EMPTY_FILTER` as a default arg, we don't want it to be accidentally modified. Can we make this class a named tuple? I think this wouldn't require any changes at the call site, except that you can't modify the fields after you create an instance (which I don't think we do?)

gnosek (dismissed):

gnosek (approved):

PR #69, opened 2023-01-24
propagate build results and produce a nice table

iurly:
* Add exception handling in futures

* Propagate build results for each built kernel, and produce a nice textual table with two results for each kernel built: one for kmod and one for ebpf.

* Also add a second table with failed kernels only, so they become easier to spot

* Set the exit code to 1 in case of failures

### DIFF
--- a/probe_builder/__init__.py
+++ b/probe_builder/__init__.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import sys
+import traceback
 
 import click
 
@@ -137,8 +138,53 @@ def build(builder_image_prefix,
             future = executor.submit(distro_builder.build_kernel, workspace, probe, distro.builder_distro, release, target)
             kernels_futures.append((release, future))
 
+
+    print("List of analyzed kernels:")
+    fstr = "|{:<45}|{:<10}|{:<10}|"
+    l = fstr.format("Kernel", "kmod", "ebpf")
+    print("-" * len(l))
+    print(l)
+    print("-" * len(l))
+
+    failed = 0
     for release, future in kernels_futures:
-        print("Release: {}, Result: {}".format(release, future.result()))
+        try:
+            res = future.result()
+            if res.kmod_result.failed() or res.ebpf_result.failed():
+                failed += 1
+            print(fstr.format(release,
+                res.kmod_result.build_result_string(),
+                res.ebpf_result.build_result_string()))
+        except:
+            failed += 1
+            print(fstr.format(release, "EXCEPTION", "EXCEPTION"))
+
+    print("-" * len(l))
+    print("Number of kernels analyzed: {}".format(len(kernels_futures)))
+    print("")
+
+    if failed:
+        print("List of failed kernels:")
+        print("-" * len(l))
+        print(l)
+        print("-" * len(l))
+
+        for release, future in kernels_futures:
+            try:
+                res = future.result()
+                if res.kmod_result.failed() or res.ebpf_result.failed():
+                    print(fstr.format(release,
+                        res.kmod_result.build_result_string(),
+                        res.ebpf_result.build_result_string()))
+            except:
+                print(fstr.format(release, "EXCEPTION", "EXCEPTION"))
+                traceback.print_exc()
+
+        print("-" * len(l))
+        print("Number of failed kernels: {}".format(failed))
+        print("")
+
+    sys.exit(1 if failed else 0)
 
 
 @click.command()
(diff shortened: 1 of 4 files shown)


probe_builder/builder/distro/base_builder.py
```diff
@@ -51,38 +77,39 @@ def build_kernel_impl(cls, config_hash, container_name, image_name, kernel_dir,
...
-            lines = builder_image.run(workspace, probe, kernel_dir, release, config_hash, container_name, image_name, args)
-        except subprocess.CalledProcessError:
+            stdout = builder_image.run(workspace, probe, kernel_dir, release, config_hash, container_name, image_name, args)
+        except subprocess.CalledProcessError as e:
             took = time.time() - ts0
             logger.error("Build failed for {} probe {}-{} (took {:.3f}s)".format(label, release, config_hash, took))
+            return cls.ProbeBuildResult(cls.ProbeBuildResult.BUILD_FAILED, took, e.output)
         else:
             took = time.time() - ts0
-            output_dir = workspace.subdir('output')
```
gnosek, line 67: Why do we delete this?

probe_builder/spawn.py
```diff
@@ -20,12 +20,12 @@ def pipe(cmd, silence_errors=False, cwd=None):
...
         for line in stdout.splitlines(False):
             logger.warn(make_string(line))
-        raise subprocess.CalledProcessError(child.returncode, cmd_string)
+        raise subprocess.CalledProcessError(child.returncode, cmd_string, stdout)
     else:
         lines = stdout.splitlines(False)
         for line in lines:
             logger.debug(make_string(line))
-        return lines
+        return make_string(stdout)
```
gnosek: Hmm this change looks like a no-op when combined with the one in docker.py (above). Am I missing something?

gnosek (commented):
Minor nits, but overall it's beautiful 🤩

probe_builder/builder/distro/base_builder.py
iurly: it got moved to the top (line 80 above) because it is needed by `probe_built()` at the very beginning to check if the probe already exists and provide a consistent `ProbeBuildResult`

probe_builder/spawn.py
iurly: could be. As a matter of fact, I find the whole `make_string()` very confusing indeed.

probe_builder/builder/distro/base_builder.py
gnosek: Gotcha. I missed it, thanks!

probe_builder/spawn.py
gnosek: it's us trying to make the probe builder work both on py2 and py3, so I suppose it's obsolete by now and we could clean up some

probe_builder/builder/distro/base_builder.py
```diff
@@ -23,6 +23,32 @@ def to_s(s):
 
 
 class DistroBuilder(object):
+    class ProbeBuildResult(object):
```
deepskyblue86, line 26: can we use enum.Enum?

probe_builder/__init__.py
```diff
@@ -137,8 +138,53 @@ def build(builder_image_prefix,
...
+    print("-" * len(l))
+    print(l)
+    print("-" * len(l))
+
+    failed = 0
     for release, future in kernels_futures:
-        print("Release: {}, Result: {}".format(release, future.result()))
+        try:
+            res = future.result()
+            if res.kmod_result.failed() or res.ebpf_result.failed():
```
deepskyblue86: KernelBuildResult method?

deepskyblue86 (dismissed):

probe_builder/__init__.py
```diff
@@ -137,8 +138,53 @@ def build(builder_image_prefix,
...
+                failed += 1
+            print(fstr.format(release,
+                res.kmod_result.build_result_string(),
+                res.ebpf_result.build_result_string()))
+        except:
+            failed += 1
+            print(fstr.format(release, "EXCEPTION", "EXCEPTION"))
+
+    print("-" * len(l))
+    print("Number of kernels analyzed: {}".format(len(kernels_futures)))
```
francesco-racciatti, line 163: w/ python3 we can use `f""` strings

iurly: Just because _you can_, it doesn't mean _you should_.
I just don't like f"" strings -- burying a variable name in a quoted string just doesn't feel right, at least to me.
Plus, they were introduced with python 3.6.

gnosek (approved):

PR #70, opened 2023-01-26
fix(debian): fix building with newer package versions

iurly:
It's not uncommon for debian to release newer package versions for the exact same kernel release, for instance:

- linux-headers-5.10.0-20-amd64_5.10.158-1_amd64.deb
- linux-headers-5.10.0-20-amd64_5.10.158-2_amd64.deb

In such cases we end up with the new package (which effectively replaces the old one) to be unpacked
in the exact same location of the previous one.

Right now we're monkey-patching Makefile's in order to fix up the include path and get the build to work, using a Makefile.sysdig-orig guard file.

So when a new package version is released, the patched Makefile gets overwritten, but the guard file does not get deleted.

In order to fix this behavior, we can simply read the Makefile and try to perform a slightly different kind of patching -- only re-writing the file if there was an "include /usr/src" to be patched in the first place.

### DIFF
--- a/probe_builder/builder/distro/debian.py
+++ b/probe_builder/builder/distro/debian.py
@@ -70,15 +70,17 @@ class DebianBuilder(DistroBuilder):
             self._reparent_link(base_path, release, 'source')
 
             makefile = os.path.join(kerneldir, 'Makefile')
-            makefile_orig = makefile + '.sysdig-orig'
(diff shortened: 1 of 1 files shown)


dark-vex (approved):

francesco-racciatti (approved):
It makes sense
LGTM

gnosek (approved):
🚢

PR #73, opened 2023-01-27
fix(amazonlinux2): skip first 5.10.x kernels without wrapper

iurly:
Apparently, 5.10.x kernels shipped with AmazonLinux2 are compiled with gcc-10 which is NOT the default compiler.
Starting with 5.10.50, they started creating a
wrapper /usr/src/kernels/*/Makefile containing the following:

GCC_VER=gcc10-
CROSS_COMPILE=$(GCC_VER)
HOSTCC=$(GCC_VER)gcc
HOSTCXX=$(GCC_VER)g++
CC=$(GCC_VER)gcc
LD=$(GCC_VER)ld.bfd

which would then call the renamed Makefile->Makefile.kernel

But before then, only the standard Makefile was used, which would then use the default toolchain (gcc7.3), leading to failure.

Since it's just 4 kernels which are quite outdated, and they probably never built in the first place,
we're just going to ditch them.

### DIFF
--- a/probe_builder/builder/builder_image.py
+++ b/probe_builder/builder/builder_image.py
@@ -76,6 +76,28 @@ SKIPPED_KERNELS = [
     ("5.8.0-1023-aws", "3f7746be1bef4c3f68f5465d8453fa4d"),
 ]
 
+
+# These AmazonLinux2 kernels were built with gcc 10.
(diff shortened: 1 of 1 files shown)


francesco-racciatti (approved):
LGTM

PR #74, opened 2023-02-02
add flag to allow overwriting of probes

iurly:
For obvious optimization reasons, we're currently skipping building probes in case the required object file already exists. However, there might be cases where we need to rebuild probes because we e.g. made changes to the Dockerfiles.

Add a '-o/--overwrite' flag for this purpose.


gnosek (approved):

matteo-porta (approved):
lgtm

PR #75, opened 2023-02-02
ubuntu: enforce usage of clang-14

iurly:
Ubuntu (and many other distros) have recently started deploying clang-15 as the default clang.
Since the resulting probes might not pass the verifier, we can just enforce the usage of clang-14 when prebuiling probes.

### DIFF
--- a/Dockerfile.ubuntu-gcc11.2-bpf
+++ b/Dockerfile.ubuntu-gcc11.2-bpf
@@ -11,8 +11,8 @@ RUN echo 'deb http://archive.ubuntu.com/ubuntu/ jammy-proposed restricted main m
 		libelf-dev \
 		make \
 		pkg-config \
-		clang \
-		llvm \
+		clang-14 \
+		llvm-14 \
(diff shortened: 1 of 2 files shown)


deepskyblue86 (approved):
Let's hope it keeps working for a while 😅

gnosek (approved):

nathan-b:
I thought we already had a fix for the clang-15 issue? Is this just a workaround until the fix can make its way into the agent libs? Or is there some longer-term reason why we want to restrict clang-15?

iurly:
> I thought we already had a fix for the clang-15 issue? 

No permanent fix that I'm aware of.
At any rate, I don't think it will make it into the upcoming release.

> Is this just a workaround until the fix can make its way into the agent libs?

Even if we fix agent-libs, we'll still have to build older versions.

>Or is there some longer-term reason why we want to restrict clang-15?

Definitely needed for now, and until we stop building probes for all the affected versions.
So I'm afraid this is patch is here to stay.

PR #78, opened 2023-02-23
[wip] sign kernel modules

iurly:
This is a WIP/Proof-of-Concept change to introduce kernel module signing.
The actual signing is performed by the in-tree script sign-file.
The private key and the certificate must be provided in the same directory as the libs/ checkout, under a specially created directory signed-key/, i.e.

 + agent-libs
   + cmake
   + driver
   + proposals
   + signing-key   <-----
   + test
   + userspace

and must be called:
   + key.priv (private key in PEM format)
   + key.der  (certificate in DER format)

NOTE: this was done in order not to create any additional paths to mount as volume, (first to the outer container, and then to the specific container).

ALSO NOTE: on newer kernels the sign-file would be a C program which also accepts a PEM textual file for the certificate but older kernels use a perl script which can only read a DER binary file, so for compatibility reason this _must_ be a binary file.

The top-level python script must be called with the '-x/--sign' flag in order for module signing to be enabled.
Since different kernels might use a different hashing algorithm (sha256, sha512, etc..), the one chosen by the target kernel is read from the kernel config file under CONFIG_MODULE_SIG_HASH.

This value is then passed over to the downstream builder as SIGN_FILE_HASH_ALGO.
When '-x/--sign' is not specified, this is left empty and the module is NOT signed.

### DIFF
--- a/Dockerfile.centos-gcc4.8
+++ b/Dockerfile.centos-gcc4.8
@@ -10,6 +10,7 @@ RUN yum -y install \
 	flex \
 	make \
 	cmake \
+	openssl \
 	elfutils-devel \
 	python-lxml && yum clean all
 
--- a/builder-entrypoint.sh
+++ b/builder-entrypoint.sh
@@ -9,6 +9,7 @@
 # PROBE_DEVICE_NAME
 # PROBE_NAME
 # PROBE_VERSION
+# SIGN_FILE_HASH_ALGO
 
 # optional env vars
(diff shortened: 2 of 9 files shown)


gnosek (dismissed):
I'm not 100% thrilled about adding stuff to the agent-libs checkout but I see why you went that way.

LGTM!

nathan-b (approved):

PR #79, opened 2023-02-27
fix(ubuntu,jammy): fix usage of clang-14

iurly:
With PR #75 we enforced usage of clang-14 for
ubuntu jammy and kinetic builders.
However, we forgot to add the explicit variables
for jammy so all builds are bound to fail.
Fix it.


nathan-b (approved):

francesco-racciatti (approved):
Thanks for the live explanation!

PR #83, opened 2023-03-09
amazonlinux: add support for AmazonLinux2023

iurly:


probe_builder/kernel_crawler/amazonlinux.py
```diff
@@ -76,14 +76,27 @@ class AmazonLinux2022Mirror(repo.Distro):
...
+
+    # cat /etc/yum.repos.d/amazonlinux.repo
+    #[amazonlinux]
+    #name=Amazon Linux 2023 repository
+    #mirrorlist=https://cdn.amazonlinux.com/al2023/core/mirrors/$releasever/$basearch/mirror.list
+
+    AL202X_BASE_URLS = ["https://cdn.amazonlinux.com/al2022", "https://cdn.amazonlinux.com/al2023"]
 
     def __init__(self):
         super(AmazonLinux2022Mirror, self).__init__([])
```
dark-vex, line 89: @iurly What do you think about changing the [class](https://github.com/draios/probe-builder/blob/dev/probe_builder/kernel_crawler/amazonlinux.py#L64) + [import](https://github.com/draios/probe-builder/blob/dev/probe_builder/kernel_crawler/__init__.py#L2) to `AmazonLinux202XMirror`?

dark-vex: and also on this [file](https://github.com/draios/probe-builder/blob/dev/probe_builder/__init__.py#L81) if we go with a more generic name for AmazonLinux

iurly: @dark-vex I thought about that, but this essentially means also changing all calling jobs.
Not quite sure it's worth it.

dark-vex (approved):

PR #84, opened 2023-03-09
crawl repositories in parallel

iurly:
this is particularly useful with centos,
whose repositories are particularly slow


francesco-racciatti (approved):
Makes sense

PR #85, opened 2023-03-09
crawling(centos): move from cern.ch to archive.kernel.org

iurly:
we've been abusing CERN's servers for a long time. 
archive.kernel.org seems more appropriate and probably has lots of closer mirrors.


francesco-racciatti (approved):
That's great.
It was a big con**CERN** to me, too!

PR #88, opened 2023-03-21
fix(debian): fix patching Makefiles, take 2

iurly:
With #70 we changed the way we patch debian Makefiles. However, the wrong assumption was made that all Makefile's would have an "include /usr.." line, whereas in reality there is also a second way:

MAKEARGS := -C /usr/src/linux-headers-4.19.0-20-common O=/usr/src/linux-headers-4.19.0-20-arm64

So adapt the patching mechanism accordingly.

### DIFF
--- a/probe_builder/builder/distro/debian.py
+++ b/probe_builder/builder/distro/debian.py
@@ -77,7 +77,11 @@ class DebianBuilder(DistroBuilder):
             target_in_container = target.replace(workspace.workspace, '/build/probe')
             with open(makefile) as fp:
                 orig = fp.read()
-            patched = orig.replace('include /usr/src', 'include ' + os.path.join(target_in_container, 'usr/src'))
+            patched = orig
+            newpath = os.path.join(target_in_container, 'usr/src')
+            patched = patched.replace('include /usr/src', 'include ' + newpath)
+            patched = patched.replace('-C /usr/src', '-C ' + newpath)
+            patched = patched.replace('O=/usr/src', 'O=' + newpath)
             if patched != orig:
                 with open(makefile, 'w') as fp:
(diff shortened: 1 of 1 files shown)


probe_builder/builder/distro/debian.py
```diff
@@ -77,7 +77,11 @@ def unpack_kernels(self, workspace, distro, kernels):
             target_in_container = target.replace(workspace.workspace, '/build/probe')
             with open(makefile) as fp:
                 orig = fp.read()
-            patched = orig.replace('include /usr/src', 'include ' + os.path.join(target_in_container, 'usr/src'))
+            patched = orig
+            newpath = os.path.join(target_in_container, 'usr/src')
+            patched = patched.replace('include /usr/src', 'include ' + newpath)
+            patched = patched.replace('-C /usr/src', '-C ' + newpath)
+            patched = patched.replace('O=/usr/src', 'O=' + newpath)
```
francesco-racciatti, line 84: Dumb question.
What if?
```python
patched = patched.replace('/usr/src', os.path.join(target_in_container, 'usr/src'))
```

iurly: because that would NOT be idempotent -- the replacement string would be `/new/path/usr/src...` so the next time around you would get `/new/path/new/path/usr/src...`.
We might get away with a regex that catches all of the word separators (`^`, ` `, `=`,...) but I got too lazy to go down that path.

francesco-racciatti:
```
usr/src
```
That's right 🤦

francesco-racciatti (approved):

dark-vex (approved):

PR #89, opened 2023-03-21
fix(download): print exceptions in futures

iurly:
we're currently downloading files by submitting download tasks to an executor, but we're not even checking whether the future results in an exception (a common use case would be a full disk). So check the future result and dump the exception if that is the case.


francesco-racciatti (approved):
LGTM

PR #91, opened 2023-03-21
feat(artifactory): add arch-based filtering when downloading

iurly:
Add an extra (optional) parameter to artifactory commands so to be able to specify the architecture (machine) of the RPM packages to download.
If not specified, the default value will be the one of the machine the script is running from.
Notice this change will have to be merged before we can start pushing packages for other architectures onto the same repository.


francesco-racciatti (approved):
LGTM

PR #92, opened 2023-03-21
chore(deb): stop printing the entire list of packages

iurly:
such list is huge and mostly unneded. Drop it.


francesco-racciatti (approved):
I like the smell of commented-out code in the morning

PR #93, opened 2023-03-21
fix(flatcar): fix builder

iurly:
with #41 we add kernel filtering, but poor flatcar guy got completely forgotten so it's been broken ever since. 
Stop at once the discrimination against it and fix it so it can rest in peace and keep being ignored but in a working state.

### DIFF
--- a/probe_builder/builder/distro/flatcar.py
+++ b/probe_builder/builder/distro/flatcar.py
@@ -69,7 +69,7 @@ class FlatcarBuilder(DistroBuilder):
         else:
             logger.info("Build for {} probe {}-{} ({}) successful".format(label, coreos_kernel_release, config_hash, release))
 
-    def crawl(self, workspace, distro, crawler_distro, download_config=None):
+    def crawl(self, workspace, distro, crawler_distro, download_config=None, filter=''):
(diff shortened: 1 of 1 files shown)


probe_builder/builder/distro/flatcar.py
```diff
@@ -69,7 +69,7 @@ def build_kernel_impl(cls, config_hash, container_name, image_name, kernel_dir,
         else:
             logger.info("Build for {} probe {}-{} ({}) successful".format(label, coreos_kernel_release, config_hash, release))
 
-    def crawl(self, workspace, distro, crawler_distro, download_config=None):
+    def crawl(self, workspace, distro, crawler_distro, download_config=None, filter=''):
         kernels = crawl_kernels(crawler_distro)
```
francesco-racciatti: Shouldn't we pass the filter to `crawl_kernels`, too?
```suggestion
        kernels = crawl_kernels(crawler_distro, filter)
```

iurly: the whole thing is going to be nuked anyway, but good point thanks!

francesco-racciatti (approved):
LGTM

PR #99, opened 2023-06-01
allow for driver sources coming from agent-kmodule.tgz

iurly:
The current procedure is based on the assumption of having a full checkout of agent-libs.
However, the source code (for both kernel module and eBPF probe) is also available as the draios-agent-kmodule.tar.gz file where all parameters are already populated.
Enable that scenario while also retaining compatibility with the existing one.
Notice the unavoidable tweaks since there is no longer a driver/ directory available.

### DIFF
--- a/builder-entrypoint.sh
+++ b/builder-entrypoint.sh
@@ -23,6 +23,14 @@ ARCH=$(uname -m)
 
 call_cmake() {
 	SRC_DIR=$1
+
+	# Make sure we've been passed a full checkout of libs,
+	# (on which we're forced to be running cmake to configure)
+	# as opposed to an already-configured source package
+	if [[ ! -f ${SRC_DIR}/CMakeLists.txt ]]; then
+		return 1
+	fi
+
 	PROBE_NAME_PARAM=PROBE_NAME
 	PROBE_VERSION_PARAM=PROBE_VERSION
 	PROBE_DEVICE_NAME_PARAM=PROBE_DEVICE_NAME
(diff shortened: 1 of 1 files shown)


gnosek (dismissed):
nice!

builder-entrypoint.sh
```diff
@@ -23,6 +23,14 @@ ARCH=$(uname -m)
 
 call_cmake() {
 	SRC_DIR=$1
+
+	# Make sure we've been passed a full checkout of libs,
+	# (on which we're forced to be running cmake to configure)
+	# as opposed to an already-configured source package
+	if [[ ! -f ${SRC_DIR}/CMakeLists.txt ]]; then
```
deepskyblue86: Do we have any symlink case? If so, I'd better go with `-e`

deepskyblue86 (dismissed):

nathan-b (dismissed):
:heart:

mavimo (commented):
@iurly have you already checked if we don't need to edit `builder-entrypoint-coreos.sh` as well?

iurly:
> @iurly have you already checked if we don't need to edit `builder-entrypoint-coreos.sh` as well?

Hmm... you're making a good point but it looks like that file is only used by `Dockerfile.coreos-old` which in turn seems completely unused. So maybe we could just remove them both? @gnosek can you please shed some light on this?

builder-entrypoint.sh
iurly: OK done!

iurly:
> @iurly have you already checked if we don't need to edit `builder-entrypoint-coreos.sh` as well?

CoreOS is long gone. Removing the remnants with https://github.com/draios/probe-builder/pull/100

gnosek (approved):

PR #102, opened 2023-07-06
ubuntu: pin gcc to 11.2 and 11.3 on jammy builders

iurly:
gcc-11 11.3 is the required compiler for 6.2.0-25
(backported to jammy), but we currently only
have a builder marked for gcc 11.2.
As a matter of fact, this results in the builder
for gcc12.2 being picked as the next best choice,
which fails because there is no gcc-11 binary.

As a matter of fact, after the release of 22.04.2, gcc-11 will resolve to 11.3.0, which breaks the
promise of having a certain gcc version in the
dockerfile name.

So pin the gcc-11 version so that we can get two
separate builders for gcc 11.2 and 11.3,
whose version is guaranteed to be honored
even in case of future updates.


AlbertoBarba (approved):

deepskyblue86 (approved):
Pair-programmed that and it's working, let's :ship: it

PR #104, opened 2023-07-28
[SMAGENT-5267] Fix prebuilt builders for air-gapped environments

iurly:
When introducing support for cross-architectural builds, we had to add a suffix specifying the architecture (e.g. ubuntu-gcc11.2-bpf-amd64) in order to tell the multiple builders apart.

The step that prebuilds all the builders to be imported into the air-gapped host (-P) is currently implemented within the bash entrypoint and does not add that suffix, thereby invalidating the whole procedure.

So defer the actual prebuilding to a newly-introduced 'prebuild' command of the python code, so that the whole naming logic is encapsulated within the same code for both producer (prebuild) and consumer (actual building).

### DIFF
--- a/main-builder-entrypoint.sh
+++ b/main-builder-entrypoint.sh
@@ -77,10 +77,10 @@ build_probes()
 	probe_builder ${DEBUG_PREFIX} build -s /sysdig -b "$BUILDER_IMAGE_PREFIX" "$@" /kernels/*
 }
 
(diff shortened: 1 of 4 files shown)


dark-vex (approved):
I did a test and it's working fine now, thanks @iurly !

mavimo (approved):

PR #111, opened 2023-10-02
feat: introduce drel (distro release)

iurly:
This PR changes the logic used by the crawler and then propagated to the builder so that the "release" (previously just the kernel release) is now a tuple composed of two parts:
- drel (distro release, e.g. 'kinetic')
- kernel (kernel release, same as before)

This brings in two significant advantages:
1. Fixes the issue with backported debian kernels, in that we only resolved packages across repositories within the same distro release (e.g. buster), see SMAGENT-5278
2. Each individual kernel build has more context (i.e. the distro release where that kernel was found). This way, when trying to reproduce that kernel build, we can also specify e.g. `-R buster` to only crawl packages within that distro release, reducing overall execution time. It also gives more context as to what image would be required to reproduce a full testing environment for that kernel.

This is done through the following changes:
- introduce Mirror.get_drel_repos() as a replacement for list_repos(), which instead of returning a bare list of Repository, will return a dict where the key is the drel.
- Particularly, implement .get_drel_repos() for DebMirror (base class for both Ubuntu and Debian) so that the (deb)dists (e.g. 'kinetic', 'kinetic-updates', ...) are all grouped under the corresponding distro release 'kinetic'.
- ditto for RpmMirror
- update Mirror.get_package_tree() so to leverage the new classification (list_drel_repos()) and provide the tuple as the key
- change the logic within DebianMirror.get_package_tree() so to adapt for the new classification returned by list_dreal_repos() AND resolve dependencies among the DebRepository-es belonging to the same drelease (previously it was a global heap of ALL repositories). NOTE: this will resolve SMAGENT-5278 i.e. the issue where backported 6.1.0 kernels will use the (overall) newest linux-kbuild package which is NOT compatibile with older Debian releases.
- update {CentosBuilder,UbuntuBuilder,DebianBuilder}.unpack_kernels() so to expect a tuple as the key of the input 'kernels' parameter
- redefine single-mirror distros (i.e. AmazonLinux, Oracle, PhotonOs) so to be a subclass of repo.Mirror instead of repo.Distro. This way we don't fall for the trap of Distro.list_drel_repos() calling self.get_mirrors() which is indeed (and rightfully so) not implemented.
- refactor the code of probe_builder.build() so to print an extra column (i.e. the distro release)

### DIFF
--- a/probe_builder/kernel_crawler/repo.py
+++ b/probe_builder/kernel_crawler/repo.py
@@ -3,8 +3,11 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from collections import namedtuple
 import click
+import logging
 import os
 import sys
+
+logger = logging.getLogger(__name__)
 
 def machine2arch(mach):
     mach2arch = {
@@ -16,6 +19,15 @@ def machine2arch(mach):
 CrawlerFilter = namedtuple("CrawlerFilter", ["machine", "arch", "distro_filter", "kernel_filter"], defaults=[os.uname().machine, machine2arch(os.uname().machine), "", ""])
 
 EMPTY_FILTER=CrawlerFilter()
+
+# A repo.Repository is a collection of packages, implmemented through .get_package_tree()
+#
+# A repo.Mirror is a collection of repositories, returned by .list_repos()
+#  TODO: should the Mirror be a subclass of Repository since it also implements get_package_tree?
+
+# A repo.Distro is a collection of mirrors (and a subclass of mirror, so to be a drop-in replacement),
+#   returned by .get_mirrors()
+#   for which the list of repositories is just the union of all the repositories of each mirror
 
 class Repository(object):
     def get_package_tree(self, crawler_filter):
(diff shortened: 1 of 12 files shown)


probe_builder/kernel_crawler/repo.py
```diff
@@ -17,6 +20,15 @@ def machine2arch(mach):
 
 EMPTY_FILTER=CrawlerFilter()
 
+# A repo.Repository is a collection of packages, implmemented through .get_package_tree()
```
deepskyblue86:
```suggestion
# A repo.Repository is a collection of packages, implemented through .get_package_tree()
```

deepskyblue86 (dismissed):
LGTM

aroberts87 (dismissed):
lgtm! as discussed on our call earlier, we can coordinate some pairing sessions to make broader improvements over time outside of the scope of this PR as well.

aroberts87 (approved):
re-approving!

deepskyblue86 (approved):

PR #112, opened 2023-10-16
feat: introduce driver/kernel yaml ignorelist

iurly:
introduce an external ignorelist, i.e. a way to specify a set of kernel releases that should be skipped for a given set of driver versions.
The set of releases is specified through a regex and a conditional expression on the exploded values, expressed as a Jinja2 template.
The most notable use case is an older driver version which is incompatible with newer kernels due to compiler issues.


probe_builder/builder/blacklist.py
```diff
@@ -0,0 +1,86 @@
...
+                self.matchers[mn] = mv
+            self.blacklist = config['blacklists']
+        self.probe_version = probe_version
+
+    def blacklist_reason(self, probe_kind, kernel_release):
+        logger.debug("======== {} probe {} kernel {}".format(probe_kind, self.probe_version, kernel_release))
+        for b in self.blacklist:
+            logger.debug("==== evaluating blacklist entry {}".format(b['description']))
+            if self.probe_version in b['probe_versions']:
+                logger.debug("probe version {} matches one of agent versions {}".format(self.probe_version, b['probe_versions']))
```
gnosek: This could be info level

probe_builder/builder/blacklist.py
```diff
@@ -0,0 +1,86 @@
...
+            else:
+                logger.debug("probe version {} NOT in {}".format(self.probe_version, b['probe_versions']))
+                continue
+
+            if not b['probe_kinds'] or probe_kind in b['probe_kinds']:
+                logger.debug("probe kind {} matches one of {}".format(probe_kind, b['probe_kinds']))
```
gnosek: Same, could be info level

probe_builder/builder/blacklist.py
```diff
@@ -0,0 +1,86 @@
...
+
+            m = self.matchers[b['matcher']].search(kernel_release)
+            if m:
+                logger.debug("kernel {} matches matcher {}".format(kernel_release, b['matcher']))
+                logger.debug("unpack result: {}".format(m.groupdict()))
+                sif = b['skip_if']
+                t = jinja2.Template(sif)
+                r = t.render(m.groupdict())
+                if r and r != "False":
+                    logger.debug("== Blacklisted by {}, eval returned {}({})".format(b['description'], r, type(r)))
```
gnosek: Same, could be info level

probe_builder/builder/blacklist.py
```diff
@@ -0,0 +1,86 @@
...
+        if yamldoc:
+            config = yaml.safe_load(yamldoc)
+            for mn, mv in config['matchers'].items():
+                mv = re.compile(mv)
+                self.matchers[mn] = mv
+            self.blacklist = config['blacklists']
+        self.probe_version = probe_version
+
+    def blacklist_reason(self, probe_kind, kernel_release):
+        logger.debug("======== {} probe {} kernel {}".format(probe_kind, self.probe_version, kernel_release))
```
gnosek: We could add a note that this is checking the blacklist, make this info level and drop the logs in build_kernel()

probe_builder/builder/blacklist.py
```diff
@@ -0,0 +1,86 @@
...
+    handler.setFormatter(logging.Formatter('%(asctime)s %(message)s'))
+    handler.setLevel(logging.DEBUG)
+    logger.addHandler(handler)
+    kbl = KernelBlacklist(_test_blacklist, "12.16.3")
+    assert kbl.blacklist_reason("kmod", "4.18.0-497.el8.x86_64")
+    assert not kbl.blacklist_reason("kmod", "4.18.0-506.el8.x86_64")
+    assert not kbl.blacklist_reason("kmod", "6.2.0")
+    kbl = KernelBlacklist(_test_blacklist, "12.12.0")
+    assert kbl.blacklist_reason("kmod", "6.2.0")
+    assert kbl.blacklist_reason("kmod", "7.0.0")
```
gnosek: Ideally we'd check the exact reason as well (not that it's just not empty) to serve as documentation

gnosek (dismissed):
Some nits about logging, but LGTM!

probe_builder/builder/blacklist.schema.json
```diff
@@ -0,0 +1,66 @@
+{
+  "title": "Blacklist",
+  "type": "object",
+  "properties": {
+    "matchers": {
+      "type": "object",
+      "description": "list of matches",
+      "default": []
```
mavimo:
```suggestion
      "patternProperties": {
        "^[a-z_]*$": {
          "type": "string",
        }
      },
      "default": []
```

probe_builder/builder/blacklist.py
```diff
@@ -0,0 +1,86 @@
...
+                    return b['description']
+            else:
+                logger.debug("kernel {} does NOT belong to matcher {}".format(kernel_release, b['matcher']))
+
+        logger.debug("== kernel NOT blacklisted")
+        return None
+
+
+## =========== Test code =============
+_test_blacklist = """\
```
mavimo: Not sure to have test data as part of the application instead of having a test suite running out of the main application.

agent_blacklist.yaml
```diff
@@ -0,0 +1,17 @@
+# Blacklist for known versions of sysdig agent (sysdigcloud-probe)
```
mavimo: Is this file supposed to be here? 🤔

probe_builder/builder/blacklist.py
iurly: Hmm... not convinced. In the end this line will be executed for ALL kernels we'd building for that probe version. That's a little too much IMHO.

probe_builder/builder/blacklist.py
iurly: same here, isn't it too much?

probe_builder/builder/blacklist.py
iurly: well this one would indeed make sense, even though the same information is also available at the caller side

probe_builder/builder/blacklist.py
iurly: not sure why you'd want to pollute the logs with so much verbosity

probe_builder/builder/blacklist.py
iurly: absolutely agree! done in follow-up commit

probe_builder/builder/blacklist.schema.json
iurly: sorry, I missed your comments entirely. I will apply the suggestion in a follow-up commit.

probe_builder/builder/blacklist.py
iurly: you're right, but since there is no existing test infrastructure to hook up into I opted for the "quick and dirty" way to at least have a way to test (albeit manually)

agent_blacklist.yaml
iurly: It looked like the quickest way to get this going (and share it with customers if necessary, too)

probe_builder/builder/blacklist.schema.json
iurly: done in c193cfa

probe_builder/__init__.py
```diff
@@ -156,7 +163,7 @@ def build(builder_image_prefix,
         kernels_futures = []
         for release, target in kernel_dirs:
             drel, krel = release if type(release) is tuple else ("", release)
-            future = executor.submit(distro_builder.build_kernel, workspace, probe, distro.builder_distro, krel, target)
+            future = executor.submit(distro_builder.build_kernel, kil, workspace,  probe, distro.builder_distro, krel, target)
```
aroberts87:
```suggestion
            future = executor.submit(distro_builder.build_kernel, kil, workspace, probe, distro.builder_distro, krel, target)
```

probe_builder/builder/ignorelist.py
```diff
@@ -0,0 +1,87 @@
...
+
+class KernelIgnoreList:
+    def __init__(self, yamldoc, probe_version):
+        self.matchers = {}
+        self.ignorelist = []
+        if yamldoc:
+            config = yaml.safe_load(yamldoc)
+            for mn, mv in config['matchers'].items():
+                mv = re.compile(mv)
+                self.matchers[mn] = mv
```
aroberts87: `self.matchers = {mn: re.compile(mv) for mn, mv in config['matchers'].items()}`
☝🏻 this reads a bit easier to me, but you don't need to change it.

probe_builder/builder/ignorelist.py
```diff
@@ -0,0 +1,87 @@
...
+            self.ignorelist = config['ignorelist']
+        self.probe_version = probe_version
+
+    def ignore_reason(self, probe_kind, kernel_release):
+        logger.debug("======== {} probe {} kernel {}".format(probe_kind, self.probe_version, kernel_release))
+        for b in self.ignorelist:
+            logger.debug("==== evaluating ignorelist entry {}".format(b['description']))
```
aroberts87: Wouldn't we want to surround the description in quotes or something to distinguish it from the rest of the message?

aroberts87 (dismissed):
lgtm. a few minor notes that you can decide to take or leave. 🙂

probe_builder/builder/blacklist.py
gnosek: Oh. It looked like a log only for skipped kernels to me. Same for all the other logs, if they're emitted for all kernels, then yeah, that's too much, but I think we could use a log to tell us that a particular kernel is explicitly skipped because $REASON (I don't have a strong opinion on what log that is though).

probe_builder/builder/distro/base_builder.py
```diff
@@ -108,12 +108,20 @@ def build_kernel_impl(cls, config_hash, container_name, image_name, kernel_dir,
...
                 return cls.ProbeBuildResult(cls.ProbeBuildResult.BUILD_FAILED, took, stdout)
 
-    def build_kernel(self, workspace, probe, builder_distro, release, target):
+    def build_kernel(self, ignorelist, workspace, probe, builder_distro, release, target):
         config_hash = self.hash_config(release, target)
         output_dir = workspace.subdir('output')
 
         kmod_skip_reason = builder_image.skip_build(workspace.machine, probe, output_dir, release, config_hash, False)
+        if not kmod_skip_reason:
+            logger.info("Querying kmod ignorelist for {}".format(release))
```
gnosek: This looks like it's going to happen for every kernel, right? I was thinking about replacing it (and the eBPF one below) with the slightly more informative log in ignorelist.py:22

iurly: Yes, this is printed for every kernel. I used it essentially for debugging the change, so it should be more of a `logger.debug()` in my opinion.

probe_builder/builder/blacklist.py
iurly: Actually such a log is already due to be printed at later point in time: https://github.com/draios/probe-builder/pull/112/files/593918521e7e1843dcaaac4cc7d41e7326302643#diff-de0cd1a9ef33f7f8110f51f64de81f61570a645378c421e59a8074c1892f02fbR89

My take is that one should be enough but if there's any other detail I'm missing, please shoot!

probe_builder/__init__.py
iurly: done in cc847ff

probe_builder/builder/ignorelist.py
iurly: done in cc847ff

probe_builder/builder/ignorelist.py
iurly: done in cc847ff

probe_builder/builder/distro/base_builder.py
iurly: converted to `logger.debug()` in cc847ff

AlbertoBarba (dismissed):

francesco-furlan (dismissed):

mavimo (dismissed):

probe_builder/__init__.py
```diff
@@ -130,11 +130,12 @@ def prebuild(builder_image_prefix, machine):
 @click.option('-t', '--download-timeout', type=click.FLOAT)
 @click.option('-v', '--probe-version')
 @click.option('-m', '--machine', default=os.uname().machine)
+@click.option('-l', '--kernel_ignorelist', default='')
```
deepskyblue86: All the options are with dash, with this we'll have
- `--kernel-filter`
- `--kernel-type`
... and `--kernel_ignorelist`

probe_builder/builder/ignorelist.py
```diff
@@ -0,0 +1,85 @@
...
+logger = logging.getLogger(__name__)
+
+class KernelIgnoreList:
+    def __init__(self, yamldoc, probe_version):
+        self.matchers = {}
+        self.ignorelist = []
+        if yamldoc:
+            config = yaml.safe_load(yamldoc)
+            self.matchers = { mn: re.compile(mv) for mn, mv in config['matchers'].items() }
+            self.ignorelist = config['ignorelist']
```
deepskyblue86, line 16: Is it fine to rely on just the validator, or better to check the config keys here?

probe_builder/builder/ignorelist.py
```diff
@@ -0,0 +1,85 @@
...
+        logger.debug("======== {} probe {} kernel {}".format(probe_kind, self.probe_version, kernel_release))
+        for b in self.ignorelist:
+            logger.debug("==== evaluating ignorelist entry '{}'".format(b['description']))
+            if self.probe_version in b['probe_versions']:
+                logger.debug("probe version {} matches one of agent versions {}".format(self.probe_version, b['probe_versions']))
+            else:
+                logger.debug("probe version {} NOT in {}".format(self.probe_version, b['probe_versions']))
+                continue
+
+            if not b['probe_kinds'] or probe_kind in b['probe_kinds']:
```
deepskyblue86, line 29: So no probe_kinds means any kind will be considered for the ignorelist, right?

probe_builder/builder/ignorelist.py
```diff
@@ -0,0 +1,85 @@
...
+                logger.debug("probe kind {} matches one of {}".format(probe_kind, b['probe_kinds']))
+            else:
+                logger.debug("probe kind {} does NOT match one of {}".format(probe_kind, b['probe_kinds']))
+                continue
+
+            m = self.matchers[b['matcher']].search(kernel_release)
+            if m:
+                logger.debug("kernel {} matches matcher {}".format(kernel_release, b['matcher']))
```
deepskyblue86, line 37: is the kernel matching the matcher, or the matcher matching the kernel?

deepskyblue86 (dismissed):
Nice!
I think it would be useful to have a brief of your design doc as a `.md` file.

probe_builder/builder/ignorelist.py
iurly: correct

probe_builder/builder/ignorelist.py
iurly: that's a very good question... which I honestly don't want to answer

probe_builder/__init__.py
iurly: renamed as `--ignore-list`

probe_builder/builder/ignorelist.py
```diff
@@ -0,0 +1,86 @@
...
+    logger.addHandler(handler)
+    kil = KernelIgnoreList(_test_ignorelist, "12.16.3")
+    assert "492<=rhel<=500" == kil.ignore_reason("kmod", "4.18.0-497.el8.x86_64")
+    assert None == kil.ignore_reason("kmod", "4.18.0-477.27.1.el8_8.x86_64")
+    assert None == kil.ignore_reason("kmod", "4.18.0-506.el8.x86_64")
+    assert None == kil.ignore_reason("kmod", "6.2.0")
+    kil = KernelIgnoreList(_test_ignorelist, "12.12.0")
+    assert "kernel>=6.2" == kil.ignore_reason("kmod", "6.2.0")
+    assert "kernel>=6.2" == kil.ignore_reason("kmod", "7.0.0")
+    assert None == kil.ignore_reason("kmod", "6.1.99")
```
francesco-racciatti, line 86: 🙂

francesco-racciatti (approved):
lgtm

mavimo (approved):

deepskyblue86 (approved):

PR #115, opened 2023-10-18
fix(ignorelist): refine broke agents for kernel 6.3

iurly:
- actual first broken kernel was 6.3 not 6.4
- turns out ebpf was also affected, but got fixed with 12.14.1 (as opposed to kmod which only got fixed with 12.15.0)


francesco-racciatti (approved):
lgtm

PR #123, opened 2024-01-29
chore: [SMAGENT-6415] exclude broen kernel due to backport

mavimo:
## Changes
 - Exclude kernels that are affected by:
     * `./include/linux/export.h:17:22: error: passing argument 1 of 'class_create' from incompatible pointer type [-Werror=incompatible-pointer-types]`
     * `/code/sysdig-rw/main.c:2875:23: error: too many arguments to function 'class_create'`


iurly (approved):
Notice how eBPF failures have an entirely different root cause (though co-located):

```
/code/sysdig-rw/bpf/fillers.h:858:56: error: member reference base type 'struct percpu_counter[4]' is not a structure or union
```

PR #125, opened 2024-02-12
fix(sysdig-probe-loader): tolerate failing rmmod

iurly:
if the probe loader script is invoked with
SYSDIG_AGENT_DRIVER=universal_ebpf
it will try to rmmod the kernel driver,
which might potentially not have been loaded
in the first place.

Tolerate such a failure.

### DIFF
--- a/sysdig-probe-loader
+++ b/sysdig-probe-loader
@@ -764,7 +764,7 @@ if ([ "$SYSDIG_AGENT_DRIVER" = universal_ebpf ] || [ "$SYSDIG_AGENT_DRIVER" = km
   exit 1
 elif [ "$SYSDIG_AGENT_DRIVER" = universal_ebpf ]; then
   # We don't need the kmod for Universal eBPF
-  /sbin/rmmod "$PROBE_NAME"
+  /sbin/rmmod "$PROBE_NAME" || true
 # Loads the probe
 elif [ "$SYSDIG_AGENT_DRIVER" = legacy_ebpf ] || [ "${SYSDIG_BPF_PROBE+x}" = x ]; then
 	load_bpf_probe


francesco-racciatti (approved):

sysdig-probe-loader
```diff
@@ -764,7 +764,7 @@ if ([ "$SYSDIG_AGENT_DRIVER" = universal_ebpf ] || [ "$SYSDIG_AGENT_DRIVER" = km
   exit 1
 elif [ "$SYSDIG_AGENT_DRIVER" = universal_ebpf ]; then
   # We don't need the kmod for Universal eBPF
-  /sbin/rmmod "$PROBE_NAME"
+  /sbin/rmmod "$PROBE_NAME" || true
```
deepskyblue86, line 767: Should we check if it's there, instead of just ignoring the failure?
```suggestion
  lsmod | grep -q "^$PROBE_NAME " && /sbin/rmmod "$PROBE_NAME"
```

sysdig-probe-loader
```diff
@@ -764,7 +764,7 @@ if ([ "$SYSDIG_AGENT_DRIVER" = universal_ebpf ] || [ "$SYSDIG_AGENT_DRIVER" = km
   exit 1
 elif [ "$SYSDIG_AGENT_DRIVER" = universal_ebpf ]; then
   # We don't need the kmod for Universal eBPF
```
deepskyblue86, line 766: Not related to this PR, but... what about legacy_ebpf? I assume the rmmod should happen there too...

sysdig-probe-loader
mprzybylski: I would feel better about replacing this line with `remove_kernel_probe || or true` instead.  (See `remove_kernel_probe` function source [here](https://github.com/draios/probe-builder/blob/31f6ec4d72a46b3f1039f88df85b9fb9a12705cf/sysdig-probe-loader#L189))

It already implements @deepskyblue86's idea of checking for the presence of the kernel module before trying to `rmmod` it.

Also, I don't have strong feelings about whether we use `|| true` to ignore an `rmmod`/`remove_kernel_probe` failure, or not.  I will defer to your judgement there.

sysdig-probe-loader
mprzybylski: You are correct, @deepskyblue86.  @iurly and I will figure out how that gets done.

mprzybylski (approved):
It will get the job done, but if we merge it as-is, we definitely need another PR to add an `rmmod` step before trying calling `load_bpf_probe` in the legacy_ebpf case.

PR #131, opened 2024-03-28
fix(bpf): source checkout after package split

iurly:
The following changes are required to be able to build the ebpf probe from a libs checkout after packages have been split:
- we need to pass -DBUILD_BPF to cmake
- we need to invoke the `bpf` target from the top-level build directory (not in driver)

while at it, again for building against a libs checkout:
- invoke the `driver` target from the top-leven build directory (for consistency)


builder-entrypoint.sh
```diff
@@ -40,7 +40,7 @@ call_cmake() {
 		PROBE_VERSION_PARAM=DRIVER_VERSION
 		PROBE_DEVICE_NAME_PARAM=DRIVER_DEVICE_NAME
 	fi
-	cmake -DCMAKE_BUILD_TYPE=Release -D${PROBE_NAME_PARAM}=$PROBE_NAME -D${PROBE_VERSION_PARAM}=$PROBE_VERSION -D${PROBE_DEVICE_NAME_PARAM}=$PROBE_DEVICE_NAME -DCREATE_TEST_TARGETS=OFF ${SRC_DIR}
+	cmake -DBUILD_BPF=On -DCMAKE_BUILD_TYPE=Release -D${PROBE_NAME_PARAM}=$PROBE_NAME -D${PROBE_VERSION_PARAM}=$PROBE_VERSION -D${PROBE_DEVICE_NAME_PARAM}=$PROBE_DEVICE_NAME -DCREATE_TEST_TARGETS=OFF ${SRC_DIR}
```
deepskyblue86: `-DBUILD_DRIVER=On` for consistency? I'd like it better being explicit, so the reader knows we're building both probes.
```suggestion
	cmake -DBUILD_DRIVER=On -DBUILD_BPF=On -DCMAKE_BUILD_TYPE=Release -D${PROBE_NAME_PARAM}=$PROBE_NAME -D${PROBE_VERSION_PARAM}=$PROBE_VERSION -D${PROBE_DEVICE_NAME_PARAM}=$PROBE_DEVICE_NAME -DCREATE_TEST_TARGETS=OFF ${SRC_DIR}
```

deepskyblue86 (dismissed):

builder-entrypoint.sh
iurly: ok, though for the record AFAICT the only effect of that flag is to include `driver` in the `all` target

deepskyblue86 (approved):

PR #134, opened 2024-05-16
Added new debian-security path to Debian crawler

dark-vex:
Debian is releasing some kernel security update on `http://security.debian.org/debian-security/dists/` instead of `http://security.debian.org/dists/`. This PR adds this we path to the crawler


iurly (approved):
LGTM.
Particularly appreciative of the almost-symmetrical look of the new URL around '.org/'.
Too bad it's debian-security and not debian.security.

PR #139, opened 2024-06-24
Look for kernel configuration only when needed

deepskyblue86:
Before 5a5cbed, `get_kernel_config_hash` was called only when trying to find precompiled kernel probes (either locally or via download).
Restore that behavior and avoid an unwanted early failure.


dark-vex (approved):

francesco-racciatti (approved):
lgtm

iurly (approved):

PR #155, opened 2025-02-24
fix(debian): install fake pahole for kernel 6.12.16

iurly:
Starting from kernel 6.12.16, it appears pahole is required to build a kernel module, otherwise the build process will fail. Simply add it as a symlink to true.


francesco-racciatti (approved):
lgtm

PR #159, opened 2025-04-30
[SMAGENT-9129] Implement kernel cache

iurly:
Implement a rudimentary kernel cache that would easily reduce crawling
efforts. Logic is as follows.

A crawl is bound to a specific query, that is: kernel type (aka distro),
architecture, distro filter, kernel filter.

Whenever an actual crawl is successful, save it into the cache
file together with its timestamp.

Whenever a crawl is requested, look up the cache to see if a resultset
is already present with a timestamp no older than 20 hours -- just to
make sure we perform a crawl at least once a day. If one is available,
use it -- otherwise, just proceed as usual with a regular crawl and
cache the resultset to be reused at the next iteration.

### DIFF
--- a/probe_builder/__init__.py
+++ b/probe_builder/__init__.py
@@ -137,11 +137,13 @@ def prebuild(builder_image_prefix, machine):
 @click.option('-v', '--probe-version')
 @click.option('-m', '--machine', default=os.uname().machine)
 @click.option('-l', '--ignore-list', default='')
+@click.option('--use-cache/--no-use-cache', default=True, help="Use existing cache if available")
+@click.option('--update-cache/--no-update-cache', default=True, help="Update cache with new data")
 @click.argument('package', nargs=-1)
 def build(builder_image_prefix,
           download_concurrency, jobs, kernel_type, distro_filter,
           kernel_filter, probe_name, retries,
-          source_dir, download_timeout, probe_version, machine, ignore_list, package):
+          source_dir, download_timeout, probe_version, machine, ignore_list, use_cache, update_cache, package):
     workspace_dir = os.getcwd()
     builder_source = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
@@ -170,12 +172,15 @@ def build(builder_image_prefix,
     # - kernel_filter (e.g. "5.15" to only include kernels starting with 5.15)
 
     cache_file = "kernelcache.yaml"  # Notice the cache file would live in the cwd, i.e. /workspace
-    cache = kernelcache.KernelCache(cache_file)
+    cache = kernelcache.KernelCache(cache_file, load_cache=use_cache)
     cache_key = kernelcache.KernelCache.make_key(kernel_type, machine, distro_filter, kernel_filter)
-    print("Looking up {} in cache".format(cache_key))
-
-    ### ...and also make sure the entry is not too old (e.g. more than 20 hours old) to avoid using stale data
-    cache_entry = cache.get(cache_key, max_duration_hours=20)
+    cache_entry = None
+    if use_cache:
+        print("Looking up {} in cache".format(cache_key))
+        ### ...and also make sure the entry is not too old (e.g. more than 20 hours old) to avoid using stale data
+        cache_entry = cache.get(cache_key, max_duration_hours=20)
+    else:
+        print("Cache disabled, skipping cache lookup")
 
     if cache_entry is not None:
         # If we had a cache hit, use the cached kernels...
@@ -183,8 +188,11 @@ def build(builder_image_prefix,
     else:
         ### ... otherwise, perform the crawl and update the cache
         kernels = distro_obj.get_kernels(workspace, package, download_config, crawler_filter)
-        cache.put(cache_key, kernels)
-        cache.save()
+        if update_cache:
+            cache.put(cache_key, kernels)
+            cache.save()
+        else:
+            print("Cache update disabled, skipping cache update")
 
     kernel_dirs = distro_builder.unpack_kernels(workspace, distro.distro, kernels)
 
--- a/probe_builder/__init__.py
+++ b/probe_builder/__init__.py
@@ -7,7 +7,7 @@ import click
 
 from probe_builder.kernel_crawler import crawl_kernels, DISTROS
 from . import kernel_crawler, disable_ipv6, git, docker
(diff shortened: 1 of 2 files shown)


probe_builder/__init__.py
```diff
@@ -162,7 +162,18 @@ def build(builder_image_prefix,
...
     crawler_filter = kernel_crawler.repo.CrawlerFilter(machine=machine, arch=arch, distro_filter=distro_filter, kernel_filter=kernel_filter)
 
-    kernels = distro_obj.get_kernels(workspace, package, download_config, crawler_filter)
+    cache_file = "cache.yaml"
+    cache = kernelcache.KernelCache(cache_file)
+    cache_key = kernelcache.KernelCache.make_key(kernel_type, machine, arch, distro_filter, kernel_filter)
+    print("Looking up {} in cache".format(cache_key))
+
+    if cache.get(cache_key):
+        kernels = cache.get(cache_key)
```
pgcrooks-sysdig, line 171: I will be hated by almost half the world.
```suggestion
    if kernels := cache.get(cache_key):
        pass
```

probe_builder/__init__.py
```diff
@@ -162,7 +162,18 @@ def build(builder_image_prefix,
...
+    else:
+        kernels = distro_obj.get_kernels(workspace, package, download_config, crawler_filter)
+        cache.put(cache_key, kernels)
+        cache.save()
```
pgcrooks-sysdig: Not sure of the python version we have, but this could simplify:

```suggestion
    if (kernels := cache.get(cache_key)) is None:
        kernels = distro_obj.get_kernels(workspace, package, download_config, crawler_filter)
        cache.put(cache_key, kernels)
        cache.save()
```

probe_builder/builder/kernelcache.py
```diff
@@ -0,0 +1,50 @@
...
+            with open(cachefile, mode="rb") as fp:
+                yamldoc = fp.read()
+        except IOError:
+            logger.debug("cache file {} not found".format(cachefile))
+
+        if yamldoc:
+            self.cache = yaml.load(yamldoc, Loader=yaml.FullLoader)
+
+    def make_key(kernel_type, machine, arch, distro_filter, kernel_filter):
+        return "{}-{}-{}-{}-{}".format(kernel_type, machine, arch, distro_filter, kernel_filter)
```
pgcrooks-sysdig: NIT if we are running modern python; I think f-strings make these a lot easier to read.

probe_builder/builder/kernelcache.py
```diff
@@ -0,0 +1,62 @@
...
+                logger.debug("cache entry {} expired".format(key))
+                del self.cache[key]
+                return None
+        else:
+            logger.debug("cache miss for key {}".format(key))
+            return None
+
+    def put(self, key, kernels):
+        if key in self.cache:
+            logger.debug("cache entry {} already exists".format(key))
```
gnosek: the nittiest of all nits, maybe say what we're doing about it, like `updating cache entry {}` instead?

probe_builder/builder/kernelcache.py
```diff
@@ -0,0 +1,62 @@
...
+            return
+
+        try:
+            self.cache = yaml.load(yamldoc, Loader=yaml.FullLoader)
+        except yaml.YAMLError as e:
+            logger.warning("failed to parse cache file {}: {}".format(self.cachefile, e))
+
+    def save(self):
+        with open(self.cachefile, mode="w") as fp:
+            yaml.dump(self.cache, fp, default_flow_style=False)
```
gnosek: Do we want some exception handling here? I think we could survive with just ignoring write failures (it's a cache, not critical data).

Also, for extra points, we might want to write the yaml to a temporary file and then rename it into place (so that if things go wrong, we at least keep the old cache)

probe_builder/__init__.py
```diff
@@ -162,7 +162,30 @@ def build(builder_image_prefix,
...
+    # - distro_filter (e.g. "resolute" for ubuntu resolute
+    # - kernel_filter (e.g. "5.15" to only include kernels starting with 5.15)
+
+    cache_file = "kernelcache.yaml"  # Notice the cache file would live in the cwd, i.e. /workspace
+    cache = kernelcache.KernelCache(cache_file)
+    cache_key = kernelcache.KernelCache.make_key(kernel_type, machine, distro_filter, kernel_filter)
+    print("Looking up {} in cache".format(cache_key))
+
+    ### ...and also make sure the entry is not too old (e.g. more than 20 hours old) to avoid using stale data
+    cache_entry = cache.get(cache_key, max_duration_hours=20)
```
gnosek: Do we want to have an option to bypass the cache (or override the max duration, which would allow for the same thing)? Like, I don't know, emergency security patches being released?

gnosek (commented):
Yeah, LGTM, with some minor comments (feel free to take or leave them)

probe_builder/builder/kernelcache.py
iurly: excellent point, thanks!

probe_builder/__init__.py
iurly: Excellent point. And the same could be said about the `cache_file`.
It's just that adding the 14th 15th and 16th argument kinda makes me a little uncomfortable.
But having a way to bypass the cache is probably worth it -- but at that point, would we also want it to be overwritten?

probe_builder/builder/kernelcache.py
iurly: Totally agree on both points. Thanks for the suggestions!

gnosek: We're not using the temporary file, are we?

probe_builder/__init__.py
gnosek: My gut feeling is that we'd still want to update the cache (we did do a crawl after all, might as well save the results), but I don't have a strong opinion.

iurly: Yep, that's why I added two separate flags
* `--use-cache` for reaading
* `--update-cache` for writing

probe_builder/builder/kernelcache.py
iurly: nope, that felt a bit of an overkill TBH with you

gnosek (approved):

PR #165, opened 2025-07-03
fix(sysdig-probe-loader): false success reporting for eBPF download [SMAGENT-9412]

deepskyblue86:
- Introduce download_probe_common, with the already correct logic of download_kernel_probe
- Use it for both kmod and eBPF


iurly (approved):
LGTM

PR #168, opened 2025-08-04
[SMAGENT-9558] chore(rpm): debug corrupted sqlite files

iurly:
Add context (i.e. origin URL) when the sqlite file is unreadable or
corrupted.


francesco-racciatti (approved):
lgtm

PR #171, opened 2025-08-27
fix(amzn2): choose correct builder

iurly:
With #158 we introduced a dedicated builder for amzn2023 but this could also be accidentally chosen for amzn2 as it is a valid prefix. On aarch64, this will result in the following:

arch/arm64/Makefile:25: ld does not support --fix-cortex-a53-843419; kernel may be susceptible to erratum
make[2]: gcc10-gcc: No such file or directory

Fix the glob for selection.


pgcrooks-sysdig (approved):
LGTM

PR #173, opened 2025-10-28
chore(ubuntu): ignore 6.14.0-1012-nvidia

iurly:
This particular kernel seems to have (unintentially?) received a backport from 6.15, therefore breaking all builds for agent <= 14.0.0 (Boo!).

For the time being, add a specifically carved ~pumpkin~ exception for this kernel only (scary!).

We'll see if coming Halloween brings in additional tricks or treats.


pgcrooks-sysdig (approved):
LGTM

PR #174, opened 2025-11-03
chore(ubuntu): ignore 6.14.0-<x>-nvidia

iurly:
Follow up on #173.
Turns out 6.14.0-1012-nvidia was not an exception and the backport is here to stay, with a new 6.14.0-1013-nvidia knocking at our doors over the Halloween holidays to trick us into failure again. Let's treat it with respect and ignore it, too.


francesco-racciatti (approved):
lgtm

PR #177, opened 2026-01-22
chore(ubuntu): drop gcc12.2

iurly:
With #110 we somehow restored gcc12.2, using lunar. But now lunar is long gone and we can no longer get gcc12.2. So just drop it and let the system default to gcc12.3 (as suggested in #110 as the alternative).


pgcrooks-sysdig (approved):
LGTM

PR #179, opened 2026-02-16
fix(ebpf): cap at kernel 6.18 [SMAGENT-10210]

iurly:
Kernel 6.19 changed the way namespaces are handled, both internally and externally [1]. This breaks the legacy eBPF probe which is however under deprecation already, in favor of the modern/universal eBPF one. Just stop building the legacy eBPF probe for any kernel newer than 6.18.

[1] https://lwn.net/Articles/1043824/

### DIFF
--- a/probe_builder/builder/builder_image.py
+++ b/probe_builder/builder/builder_image.py
@@ -112,6 +112,12 @@ MIN_EBPF_SUPPORTED_RELEASE = {
     's390x': '5.5',
 }
 
+MAX_EBPF_SUPPORTED_RELEASE = {
+    'x86_64': '6.18',
+    'aarch64': '6.18',
+    's390x': '6.18',
+}
+
 def probe_built(mach, probe, output_dir, kernel_release, config_hash, bpf):
     probe_file_name = probe_output_file(mach, probe, kernel_release, config_hash, bpf)
     return os.path.exists(os.path.join(output_dir, probe_file_name))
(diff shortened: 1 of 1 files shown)


francesco-racciatti (approved):
lgtm

I guess we should update our docs to reflect this

probe_builder/builder/builder_image.py
```diff
@@ -112,6 +112,12 @@ def probe_output_file(mach, probe, kernel_release, config_hash, bpf):
     's390x': '5.5',
 }
 
+MAX_EBPF_SUPPORTED_RELEASE = {
```
mavimo, line 115:
```suggestion
MAX_LEGACY_EBPF_SUPPORTED_RELEASE = {
```

probe_builder/builder/builder_image.py
```diff
@@ -125,8 +131,11 @@ def skip_build(mach, probe, output_dir, kernel_release, config_hash, bpf):
     if bpf:
         kernel_version = Version(kernel_release)
         min_release = MIN_EBPF_SUPPORTED_RELEASE[mach]
+        max_release = MAX_EBPF_SUPPORTED_RELEASE[mach]
```
mavimo, line 134:
```suggestion
        max_release = MAX_LEGACY_EBPF_SUPPORTED_RELEASE[mach]
```

mavimo (approved):

probe_builder/builder/builder_image.py
iurly: I'd prefer to keep it consistent with the past and the adjacent MIN_EBPF_SUPPORTED_RELEASE. It's implicitly assumed that the probe-builder only deals with the legacy ebpf probe.

PR #181, opened 2026-02-17
fix(ubuntu): fix builder images with gcc-14

iurly:
Followup on #180
Apparently ubuntu-proposed is no longer present.
Just drop ubuntu-proposed.

Also, the default gcc-14 on ubuntu-updates is 14.2. So rename the file accordingly.


pgcrooks-sysdig (dismissed):
LGTM

pgcrooks-sysdig (approved):
LGTM

PR #182, opened 2026-03-03
fix(kernelcache): do not write an empty cache

iurly:
If kernel crawling fails, we get an empty list of kernels. This should not overwrite the existing cache, otherwise subsequent attempts will use that value and fail because of an empty list.


pgcrooks-sysdig (approved):
LGTM

PR #187, opened 2026-06-09
fix(probe-loader): attempt dkms build on UEK hosts

juan-rodriguez-gracia:
### Description

The probe loader skipped the dkms kernel-module build on any UEK kernel (`uname -r == *uek*`), under the assumption that the build would always fail. That assumption no longer holds on modern Oracle Linux / OKE nodes (e.g. OL8 + UEK6 `5.15.0-*.el8uek`), where the matching `kernel-uek-devel` sources and a compatible gcc are present on the host.

Because the build was skipped, the agent had no kmod probe, and the precompiled-probe download returns HTTP 404 for these kernels, so `slim-with-kmod` never starts (DaemonSet rollout times out).

### Change

Always attempt `build_kernel_probe`. On failure, the existing fallbacks (system module via `modprobe`, then precompiled-probe download) still run, so non-OKE UEK hosts behave exactly as before.

### Evidence

From an OKE agent pod (`kubernetes-installation #12784`, agent `0.99.13419dev`, kernel `5.15.0-320.202.8.2.el8uek.x86_64`):

```
* Will use /usr/bin/gcc-11 to build kernel module if needed
* Will use /usr/bin/gcc11-as as a dedicated assembler ...
* Loading kernel probe
* Skipping dkms install for UEK host
* Trying to download precompiled module from https://download.sysdig.com/.../...-5.15.0-320.202.8.2.el8uek.x86_64-...ko
curl: (22) The requested URL returned error: 404
Cannot load the probe
```

Kernel sources are present on the node (`/host/usr/src/kernels/5.15.0-320.202.8.2.el8uek.x86_64/`) and gcc-11 is already selected, so the on-the-fly dkms build should succeed.

### Testing

To be validated end-to-end via `kubernetes-installation` on an ephemeral OKE cluster, using an agent image built from a `draios/agent` branch that bumps this submodule.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sysdig-probe-loader
+++ b/sysdig-probe-loader
@@ -354,14 +354,13 @@ load_kernel_probe() {
 	if [ -v SYSDIG_FORCE_DOWNLOAD_PROBE ]; then
 		echo "* Skipping build, FORCE_DOWNLOAD_PROBE is enabled"
 	else
-		# Builds the probe via dkms; skips UEK hosts, the build will always fail
-		if [[ $(uname -r) == *uek* ]]; then
-			echo "* Skipping dkms install for UEK host"
-		else
-			build_kernel_probe
-			if [ $? -eq 0 ]; then
-				return 0
-			fi
+		# Build the probe via dkms. This is also attempted on UEK hosts: when
+		# the matching kernel-uek-devel sources and a compatible gcc are present
+		# (e.g. Oracle Linux / OKE nodes) the build succeeds. If it fails we fall
+		# through to loading a system module or a precompiled probe below.
+		build_kernel_probe
+		if [ $? -eq 0 ]; then
+			return 0
 		fi
 
 		echo "* Trying to load a system ${PROBE_NAME}, if present"


juan-rodriguez-gracia:
✅ Validado end-to-end en OKE efímero — `kubernetes-installation #12786` (CLUSTER_TYPE=oke, slim_with_kmod) **SUCCESS**.

Log del pod del agente (kernel `5.15.0-320.202.8.2.el8uek.x86_64`):
```
* /usr/bin/gcc-11 does not support -ftrivial-auto-var-init (required by this kernel); looking for a newer gcc
* Will use /usr/bin/gcc-12 instead to build kernel module
* Running dkms build for draios-agent
  dkms build done
* Found .../sysdigcloud-probe.ko.xz; trying to insmod
  sysdigcloud-probe.ko.xz found and loaded in dkms
Probe loaded
```
La probe kmod ahora se compila y carga al vuelo en UEK/OKE.

juan-rodriguez-gracia:
✅ Final full OKE verification passed — `kubernetes-installation #12787` (CLUSTER_TYPE=oke, ephemeral): `test_install_agent_slim_with_kmod` PASSED and `test_install_agent_slim_with_legacy_ebpf` PASSED, all agent checks green. kmod probe built on the fly with gcc-12. Ready to merge.

